PR #59, opened 2013-10-13
Support active record transaction error notification.

dtaniwaki:
Hi,

I found Bugsnag can't notify exceptions in after_commit and after_rollback because it doesn't escalate any exception. Although there were lots of exception thrown in the after_commit callbacks, but I didn't know them because I had relied on Bugsnag too much and hadn't checked the error log.
I know this is very rare case, but I think it is better if Bugsnag handles this kind of situation. What do you think?


arthurnn:
> this seems like a rails bug to me

Indeed , I fixed on rails 4.2+ .

PR https://github.com/bugsnag/bugsnag-ruby/pull/250 to update bugsnag

PR #173, opened 2014-11-07
Support operating with metadata in ignore_errors block

### LINKED ISSUE #90: Allow to notify about sidekiq error only after several insuccessful retries
We are sending emails with sidekiq workers that are monitored for exceptions by bugsnag.

There are a lot of one time exceptions like `IO::EAGAINWaitReadable`.
The 5 seconds downtime of the mail  server (e.g it is redeployed) results in hundreds of such errors.

Can we setup bugsnag-sidekiq to notify about exceptions only after several unsuccessful retries?

bogdan:
Bugfix build for ruby 1.8.

Fixes #90


ConradIrwin:
I don't think this is the right direction for this fix.

Issue #141, opened 2014-08-07
httparty is awful


pda:
Depending on httparty produces obnoxious output in `bundle install` for all apps that add `bugnag` to their Gemfile. See: https://github.com/jnunemaker/httparty/pull/321

```
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.
Post-install message from httparty:
When you HTTParty, you must party hard!
```

Would you consider a pull-request to replace httparty with plain Net::HTTP or a grown-up library like Faraday?

scarybot:
Consider using https://rubygems.org/gems/httparty-responsibly - it's a plain fork without the post-install message.

PR #183, opened 2014-12-15
Replace HTTParty with net/http

lautis:
With the new configurable delivery methods [replacing HTTParty](#141) with net/http is pretty straightforward. Less dependencies and no post-install messages!


ConradIrwin:
Thanks!

PR #187, opened 2014-12-30
Handle potential symbols as keys in the rack env

cheezy2022:
I ran into this issue when using the clearance gem for our authentication. 

In the ```rack_request.rb``` it loops thru all key/value pairs in the rack_env. One of the keys we end up having is ```:clearance```, so the String method ```starts_with?``` throws an exception because the key is a symbol, not a String.

Adding a simple ```to_s``` solves the problem.


ConradIrwin:
Thanks!

Issue #192, opened 2015-01-06
Filtering data in overrides


radar:
I have the following code in an Rails initializer:

```ruby
class ExceptionFilter
  def initialize(bugsnag)
    @bugsnag = bugsnag
  end

  def call(notification)
    sidekiq_tab = notification.meta_data[:sidekiq]
    if sidekiq_tab
      encoded_request = sidekiq_tab[:args].first[:encoded_request]
      if token = encoded_request[:token]
        notification.meta_data[:sidekiq][:args].first[:encoded_request][:token] = "#{token[0...6]}*****#{token[-4..-1]}"
      end
    end

    if token = sidekiq_tab[:args].first[:access_token]
      notification.meta_data[:sidekiq][:args].first[:access_token] = "#{token[0...6]}*****#{token[-4..-1]}"
    end

    @bugsnag.call(notification)
  end
end

Bugsnag.configure do |c|
  c.middleware.use ExceptionFilter
end
```

As you can see in the code, I am trying to sanitize the `token` and `access_token` keys which are reported in the Sidekiq tab to Bugsnag.

The issue is that this middleware is called too early, before the `meta_data` hash is generated. I've investigated this and I can see that [the middleware is called here starting on line 189](https://github.com/bugsnag/bugsnag-ruby/blob/efb8542f9c02652c34f862e1dc4206014edf3fa2/lib/bugsnag/notification.rb#L189-L204), while the generation of the `meta_data` is done later on [line 230](https://github.com/bugsnag/bugsnag-ruby/blob/efb8542f9c02652c34f862e1dc4206014edf3fa2/lib/bugsnag/notification.rb#L230).

I think the metadata should be generated earlier on and be made available to the middlewares so that they can massage it as they please. Would a patch to do this be OK?

ConradIrwin:
I think that's an excellent idea, thanks!

PR #193, opened 2015-01-06
Process meta_data before middleware stack

radar:
This is a fix for #192. Please see that issue for the background for this PR.

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -47,7 +47,9 @@ module Bugsnag
 
     # Explicitly notify of an exception
     def notify(exception, overrides=nil, request_data=nil)
-      Notification.new(exception, configuration, overrides, request_data).deliver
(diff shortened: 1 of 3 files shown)


lib/bugsnag.rb
```diff
@@ -47,7 +47,9 @@ def configure(config_hash=nil)
 
     # Explicitly notify of an exception
     def notify(exception, overrides=nil, request_data=nil)
-      Notification.new(exception, configuration, overrides, request_data).deliver
+      notification = Notification.new(exception, configuration, overrides, request_data)
+      notification.deliver
+      notification
```
radar: I needed to make this change so that I could tell in the test if `notification.meta_data` had been modified.

radar:
I am not sure how to fix that failing test. Any tips would be appreciated.

ConradIrwin:
@radar This was fixed in #205.

PR #198, opened 2015-01-14
Allow specifying heroku app for deploy hooks

loopj:


snmaynard:
I'll leave this one to @ConradIrwin to review as he has more heroku exp

ConradIrwin:
Looks good to me. `HEROKU_APP` is a reasonably well accepted env var (I think the heroku client library supports it in recent versions).

PR #200, opened 2015-01-22
Allow for a delivery failure hook

seejohnrun:
We want to hook into when synchronous delivery fails


ConradIrwin:
@seejohnrun Out of interest what do you want this hook to do?

Could you please re-send this pull request using ruby-1.8.7-compatible syntax.

PR #204, opened 2015-02-12
Support whitespace in Heroku config vars

grn:
A config var like "VAR=with space" incorrectly yields two lines:
"VAR=with" and "space". Because the latter doesn't contain an equal
sign v in the block equals nil. This results in NoMethodError raised by
v.strip.


ConradIrwin:
Thanks!

PR #205, opened 2015-02-13
Add internal_middleware stack and expose meta_data to middleware.

### LINKED ISSUE #192: Filtering data in overrides

lowensk:
Fixes #192

### DIFF
--- a/lib/bugsnag/middleware/mailman.rb
+++ b/lib/bugsnag/middleware/mailman.rb
@@ -5,7 +5,8 @@ module Bugsnag::Middleware
     end
 
     def call(notification)
-      notification.add_tab(:mailman, {"message" => notification.request_data[:mailman_msg]})
+      mailman_msg = notification.request_data[:mailman_msg]
+      notification.add_tab(:mailman, {"message" => mailman_msg}) if mailman_msg
       @bugsnag.call(notification)
     end
   end
--- a/lib/bugsnag/middleware/mailman.rb
+++ b/lib/bugsnag/middleware/mailman.rb
@@ -1,0 +1,12 @@
+module Bugsnag::Middleware
+  class Mailman
+    def initialize(bugsnag)
+      @bugsnag = bugsnag
+    end
+
+    def call(notification)
+      notification.add_tab(:mailman, {"message" => notification.request_data[:mailman_msg]})
+      @bugsnag.call(notification)
+    end
+  end
+end
--- a/lib/bugsnag/middleware/sidekiq.rb
+++ b/lib/bugsnag/middleware/sidekiq.rb
@@ -5,8 +5,8 @@ module Bugsnag::Middleware
     end
 
     def call(notification)
-      notification.add_tab(:sidekiq, notification.request_data[:sidekiq_msg])
-      notification.context ||= "sidekiq##{notification.request_data[:sidekiq_queue]}"
+      sidekiq = notification.request_data[:sidekiq]
+      notification.add_tab(:sidekiq, sidekiq) if sidekiq
       @bugsnag.call(notification)
     end
   end
--- a/lib/bugsnag/middleware/sidekiq.rb
+++ b/lib/bugsnag/middleware/sidekiq.rb
@@ -1,0 +1,13 @@
+module Bugsnag::Middleware
+  class Sidekiq
+    def initialize(bugsnag)
+      @bugsnag = bugsnag
+    end
+
+    def call(notification)
+      notification.add_tab(:sidekiq, notification.request_data[:sidekiq_msg])
+      notification.context ||= "sidekiq##{notification.request_data[:sidekiq_queue]}"
+      @bugsnag.call(notification)
+    end
+  end
+end
--- a/lib/bugsnag/notification.rb
+++ b/lib/bugsnag/notification.rb
@@ -237,12 +237,16 @@ module Bugsnag
           :exceptions => exception_list,
           :severity => self.severity,
           :groupingHash => self.grouping_hash,
-          :metaData => @meta_data
-        }.reject {|k,v| v.nil? }
+        }
 
         payload_event[:device] = {:hostname => @configuration.hostname} if @configuration.hostname
 
+        # cleanup character encodings
         payload_event = Bugsnag::Helpers.cleanup_obj_encoding(payload_event)
+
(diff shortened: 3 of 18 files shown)


spec/middleware_spec.rb
```diff
@@ -65,26 +67,41 @@
...
       event = get_event_from_payload(payload)
       expect(event["metaData"]["custom"]).not_to be_nil
-      expect(event["metaData"]["custom"]["info"]).to eq("here2")
-      expect(event["metaData"]["custom"]["data"]).to eq("also here")
+      expect(event["metaData"]["custom"]["info"]).not_to eq(InternalInfoSetter::MESSAGE)
+      expect(event["metaData"]["custom"]["info"]).to eq("overridden")
     }
   end
 
+  it "doesn't allow overrides to override public middleware" do
```
snmaynard: isnt this testing the opposite?

spec/fixtures/middleware/internal_info_setter.rb
```diff
@@ -0,0 +1,11 @@
+class InternalInfoSetter
+  MESSAGE = "set by internal_middleware"
+  def initialize(bugsnag)
+    @bugsnag = bugsnag
+  end
+
+  def call(notification)
+    notification.info = MESSAGE
```
snmaynard: I cant see how this works? Is this some ruby awesomeness I dont understand? How does this end up in metadata?

lib/bugsnag/notification.rb
```diff
@@ -235,17 +237,13 @@ def deliver
           :exceptions => exception_list,
           :severity => self.severity,
           :groupingHash => self.grouping_hash,
-        }
+          :metaData => @meta_data
+        }.reject {|k,v| v.nil? }
 
         payload_event[:device] = {:hostname => @configuration.hostname} if @configuration.hostname
 
         payload_event = Bugsnag::Helpers.cleanup_obj_encoding(payload_event)
```
ConradIrwin: @lowensk We probably want to run `cleanup_obj` on the meta-data after the user's custom handlers (they might add stupid stuff to the meta-data at that point).

lib/bugsnag/rake.rb
```diff
@@ -31,7 +18,7 @@ def bugsnag_define_task(*args, &block)
       task = self.original_define_task(*args) do |*block_args|
         begin
           old_task = Thread.current[:bugsnag_running_task]
-          Thread.current[:bugsnag_running_task] = task
+          Bugsnag.set_request_data :bugsnag_running_task, task
```
ConradIrwin: @lowensk This needs the same change in the `ensure` block.

lib/bugsnag/middleware/mailman.rb
```diff
@@ -0,0 +1,12 @@
+module Bugsnag::Middleware
+  class Mailman
+    def initialize(bugsnag)
+      @bugsnag = bugsnag
+    end
+
+    def call(notification)
+      notification.add_tab(:mailman, {"message" => notification.request_data[:mailman_msg]})
```
ConradIrwin: Should be conditional on request_data[:mailman_msg] existing.

lib/bugsnag/middleware/sidekiq.rb
```diff
@@ -0,0 +1,13 @@
+module Bugsnag::Middleware
+  class Sidekiq
+    def initialize(bugsnag)
+      @bugsnag = bugsnag
+    end
+
+    def call(notification)
+      notification.add_tab(:sidekiq, notification.request_data[:sidekiq_msg])
```
ConradIrwin: This should be conditional on the `:sidekiq_msg` actually being there.

spec/middleware_spec.rb
lowensk: You're absolutely right.

spec/fixtures/middleware/internal_info_setter.rb
lowensk: This was my lack of understanding - I was attempting to pass do...end style blocks to have_sent_notification (rather than curly brace style blocks) so these tests were passing when they shouldn't have been.

lib/bugsnag/middleware/mailman.rb
lowensk: Caught this in my local testing as well :+1:

radar:
This PR removes the require for delayed_job integration.

lowensk:
@radar are you referring to this part? https://github.com/bugsnag/bugsnag-ruby/pull/205/files?diff=split#diff-921fa4171158da4eec0fae5db27bb924L17 - that's been moved to the bottom of the file. Is there something else affecting delayed_job that I'm not seeing?

radar:
Yup that's the part. Sorry, I conflated middleware requires with integration requires.

PR #206, opened 2015-02-17
Refactor cleanup_url

### LINKED ISSUE #202: Filtering from Rails.configuration.filter_parameters does not work in URLs
In a Rails application, given the following configuration for filtering params:

```ruby
Rails.configuration.filter_parameters
=> [:password, :access_token, :refresh_token]
```

Bugsnag does not properly filter URLs that get sent across in the Request tab. In particular, given that configuration, the Bugsnag railtie [changes the Rails filters](https://github.com/bugsnag/bugsnag-ruby/blob/master/lib/bugsnag/railtie.rb#L21-L27) to strings that start with `\A` and end with `\z`, so we end up with:

```ruby
Bugsnag.configuration.params_filters
=> #<Set: {"password",
 /\Aaccess_token\z/,
 /\Arefresh_token\z/}>
```

As a result, [`Bugsnag.cleanup_url`](https://github.com/bugsnag/bugsnag-ruby/blob/master/lib/bugsnag/helpers.rb#L70-L76) does not match a URL that needs to be filtered properly:

```ruby
Bugsnag::Helpers.cleanup_url("https://host.example/sessions?access_token=abc123", Bugsnag.configuration.params_filters)
=> "https://host.example/sessions?access_token=abc123"
```

If Rails parameters are picked up unchanged (simulated by the direct assignment in the first line), `Bugsnag.cleanup_url` works as expected:

```ruby
Bugsnag.configuration.params_filters = ["password", "access_token", "refresh_token"]
Bugsnag::Helpers.cleanup_url("https://host.example/sessions?access_token=abc123", Bugsnag.configuration.params_filters)
=> "https://host.example/sessions?access_token=[FILTERED]"
```

I don’t know the motivation for modifying `Rails.configuration.filter_parameters` when including them in `Bugsnag.configuration.params_filters`, but I propose passing those values unmodified so filtering works properly everywhere.

lowensk:
Fixes #202

### DIFF
--- a/lib/bugsnag/helpers.rb
+++ b/lib/bugsnag/helpers.rb
@@ -81,6 +81,8 @@ module Bugsnag
       return url if filters.empty?
 
       uri = URI(url)
+      return url unless uri.query
+
       query_params = uri.query.split('&').map { |pair| pair.split('=') }
       query_params.map! do |key, val|
         if filters_match?(key, filters)
--- a/lib/bugsnag/helpers.rb
+++ b/lib/bugsnag/helpers.rb
@@ -1,3 +1,5 @@
+require 'uri'
+
 module Bugsnag
   module Helpers
     MAX_STRING_LENGTH = 4096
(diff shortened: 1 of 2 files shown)


lib/bugsnag/helpers.rb
```diff
@@ -75,12 +77,21 @@ def self.filters_match?(object, filters)
...
     end
 
-    def self.cleanup_url(url, filters = nil)
-      return url unless filters
+    def self.cleanup_url(url, filters = [])
+      return url if filters.empty?
 
-      filter_regex = Regexp.new("([?&](?:[^&=]*#{filters.to_a.join('|[^&=]*')}[^&=]*)=)[^&]*")
+      uri = URI(url)
+      query_params = uri.query.split('&').map { |pair| pair.split('=') }
```
ConradIrwin: @lowensk I think uri.query can be `nil`. We should guard against that.

ConradIrwin:
Otherwise looks great! :factory:

PR #208, opened 2015-02-17
Fix intermittent failure in integration_spec

lowensk:
Fixes order-dependent failure seend here:
https://travis-ci.org/bugsnag/bugsnag-ruby/jobs/51120509


ConradIrwin:
:large_orange_diamond:

PR #209, opened 2015-02-18
Move railtie params_filters configuration to after_initialize

### LINKED ISSUE #164: params_filter set too early for rails apps
According to http://guides.rubyonrails.org/configuring.html

```
config.filter_parameters used for filtering out the parameters that you don't want shown in the logs, such as passwords or credit card numbers. New applications filter out passwords by adding the following config.filter_parameters+=[:password] in config/initializers/filter_parameter_logging.rb.
```

If the initializer file is used, it will change `filter_parameters` too late for Bugsnag to pickup as the railtie uses `before_initialize` http://apidock.com/rails/v4.0.2/Rails/Railtie/Configuration/before_initialize as seen in https://github.com/bugsnag/bugsnag-ruby/blob/35920440281de38b8d2d510e2702981885c8c82b/lib/bugsnag/railtie.rb#L15

lowensk:
Fixes #164

### DIFF
--- a/lib/bugsnag/railtie.rb
+++ b/lib/bugsnag/railtie.rb
@@ -15,24 +15,16 @@ module Bugsnag
     config.before_initialize do
       # Configure bugsnag rails defaults
       Bugsnag.configure do |config|
-        config.logger = Rails.logger
-        config.release_stage = Rails.env.to_s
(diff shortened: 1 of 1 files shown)


ConradIrwin:
:mortar_board:

PR #214, opened 2015-03-04
Move building payloads, don't require a block for middleware run

BlakeMesdag:
This moves building a payload into a method for easier integration with other background queueing libraries. Also removes the requirement of a block to execute the middleware stack.

### DIFF
--- a/lib/bugsnag/middleware_stack.rb
+++ b/lib/bugsnag/middleware_stack.rb
@@ -56,7 +56,7 @@ module Bugsnag
       lambda_has_run = false
       notify_lambda = lambda do |notif|
         lambda_has_run = true
-        yield
+        yield if block_given?
       end
 
       begin
--- a/lib/bugsnag/notification.rb
+++ b/lib/bugsnag/notification.rb
@@ -191,7 +191,7 @@ module Bugsnag
       # Warn if no release_stage is set
       Bugsnag.warn "You should set your app's release_stage (see https://bugsnag.com/docs/notifiers/ruby#release_stage)." unless @configuration.release_stage
 
-      @configuration.internal_middleware.run(self) { }
+      @configuration.internal_middleware.run(self)
 
       exceptions.each do |exception|
         if exception.class.include?(Bugsnag::MetaData)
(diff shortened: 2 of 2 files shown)


ConradIrwin:
Hey @BlakeMesdag, The change looks good. Just for curiousity how are you trying to integrate with a background queueing library?

I was hoping to add a sidekiq delivery mechanism (We currently only have Delivery::Synchronous, and Delivery::ThreadPool) but that should be possible to implement without refactoring this code.

BlakeMesdag:
We have an abstraction around the actual queueing that imposes some restrictions on the kinds of things we can push to our queues. I also ran into issues serializing the Bugsnag::Configuration while a logger was set with an open IO which was throwing exceptions.

I ended up overriding the Bugsnag::Delay::Resque job to only take the first 2 parameters as a hack to try it out, but I'll be integrating it with an existing background job so we can double send to an Errbit instance for now.

Sidekiq is also API compatible with Resque, so you may not need to do a separate integration there.

ConradIrwin:
Ok, this looks good, but I'd caution against trying to call `build_exception_payload` directly. Even with your refactoring, that won't work in quite the same way. Instead you can override the delivery mechanism yourself, and then call `Bugsnag.notify`.

```ruby
Bugsnag::Delivery.register(:custom, Class.new{
  def self.deliver(url, payload, configuration)
    puts "Delivering payload to Bugsnag!"
    puts payload.to_s
    # insert clever logic here...
  end
})

# either
Bugsnag.configure do |config|
  config.delivery_method = :custom
end

#or
Bugsnag.notify(e, delivery_method: :custom)
```

BlakeMesdag:
@ConradIrwin shouldn't this build the exception properly? Or are there other external mutations applied to the notification?

```ruby
notification = Bugsnag::Notification.new(exception, Bugsnag.configuration)
payload = Bugsnag.configuration.middleware.run(notification) { build_exception_payload }
```

ConradIrwin:
Only the other stuff in [`Notification#deliver`](https://github.com/BlakeMesdag/bugsnag-ruby/blob/ccdcfe7d9b370c301365ee66b89dbd37bd27c734/lib/bugsnag/notification.rb#L180-L215), which mainly seems to be adding fields based on the arguments to `Bugsnag.notify`. You may also want to check the ignore list (but you probably already have one of those for errbit).

I don't think it's a big problem, and if you're happy to take the risk, then go for it. Where I'm coming from as maintainer/support contact is that I'd rather everyone was using the same interface. It lets me refactor things safely, and easily debug problems that people report.

BlakeMesdag:
Totally, makes sense. I'll likely use the custom handler style for the queueing then use the regular one when actually sending it. Thanks!

PR #215, opened 2015-03-11
Print a message after creating the Bugsnag initializer which warns about secrets

loopj:


ConradIrwin:
Thanks for this @loopj. I think the bugsnag token is not really secret enough to warrant this. (for bugsnag-js, the secret is actually visible over the internet).

If we were really concerned we could make the generator put the key in the right place, just telling users to do it is pretty ineffectual.

PR #216, opened 2015-03-12
Clean up strings returned from #to_s.

flexik:
When an arbitrary object included in `meta_data` hash returns a badly encoded string from its `#to_s` method, the string won't get cleaned up properly, which in turn breaks the whole notification sending.

A real life problem example is Sinatra, which stores currently thrown exception in Rack environment under `"sinatra.error"`. When that exception message contains badly encoded characters, the exception won't be reported to Bugsnag because serializing the message in `@meta_data[:environment]['sinatra_error']` fails.

The solution is to clean up all strings returned from `#to_s` methods in `Bugsnag::Helpers.cleanup_obj`.

### DIFF
--- a/lib/bugsnag/helpers.rb
+++ b/lib/bugsnag/helpers.rb
@@ -41,7 +41,7 @@ module Bugsnag
         if str =~ /#<.*>/
           '[OBJECT]'
         else
-          str
+          cleanup_string(str)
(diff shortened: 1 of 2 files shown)


ConradIrwin:
Thanks @flexik !

PR #217, opened 2015-03-23
Fix configuration thread-safety

etehtsea:
Code to reproduce:

```bash
# jruby 1.7.18
$ jruby -r'bugsnag' -e'p 128.times.map { Thread.new { Bugsnag.configuration.object_id } }.map(&:value).uniq'
[2386, 2392, 2390, 2384, 2388, 2394]
```

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -1,4 +1,5 @@
 require "rubygems"
+require "thread"
 
 require "bugsnag/version"
 require "bugsnag/configuration"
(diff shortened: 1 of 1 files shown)


ConradIrwin:
Thanks @etehtsea.

Out of interest, did you run into this problem in real life?

etehtsea:
@ConradIrwin nope. This is possible issue, bug I found it when investigated another safety issue. I'll write about it a little bit later.

PR #221, opened 2015-04-06
send notifications automatically if 'rails runner' call fails

modosc:
this change adds the suggested notification code from the 'Standard Ruby Script' section automatically to code run via ```rails runner```. 

you don't currently test for rails-specific stuff so i didn't add a spec but if it's ok to pull in all of those deps i'd  be happy to add it.


ConradIrwin:
@modosc thanks! This is great :).

I wouldn't mind having tests if you put them in development dependencies, but it seems to work manually.

PR #222, opened 2015-04-06
Don't verify if MetaData module is included at class level

mtarnovan:
This allows decorating exception instances with metadata without
extending their class with the Bugsnag::MetaData module


ConradIrwin:
Thanks! Released in v2.8.4

mtarnovan:
@ConradIrwin Appreciate the fast release, thanks.

PR #226, opened 2015-04-16
Disable sending code by default

zanker:
While the feature to send code by default is pretty cool, it causes terrible performance problems. We saw the time to send an exception go from ~1,000ms to ~200ms on a test case of:

```ruby
begin
  fail "Test"
rescue => ex
  Bugsnag.notify_or_ignore(ex)
end
```

On a lot of exceptions (such as IO), they tend to have a lot of lines because they are wrapped in a couple of exceptions, which Bugsnag unwinds and tries to load code for. It can end up being that you're opening 50+ files and making the payload for an exception much larger.

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -66,7 +66,7 @@ module Bugsnag
       self.auto_notify = true
       self.use_ssl = true
       self.send_environment = false
-      self.send_code = true
+      self.send_code = false
       self.params_filters = Set.new(DEFAULT_PARAMS_FILTERS)
       self.ignore_classes = Set.new(DEFAULT_IGNORE_CLASSES)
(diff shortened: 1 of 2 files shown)


ConradIrwin:
Going to close this now #227 is merged. In the future we'll probably also restrict code to in-project lines, would that make a significant difference for you?

zanker:
In terms of performance improvements, this one was the main one. It basically went ~1,000ms -> 200ms (this change) -> 40ms (#227).

Restricting it to in-project lines doesn't really help unfortunately. In our case (and other companies I've worked with), they bundle gems and/or Ruby, so everything is still in scope.

In terms of our code, if you were to filter out gems/Ruby, we're still looking at ~80 files being opened. If you dedup it by file, that drops to 14. It's still way too much of a performance cost for it to be enabled.

PR #227, opened 2015-04-18
Optimize exception cleanup

zanker:
Hey everyone, we found that compiling an exception (everything before `Notification#deliver_exception_payload`) takes around ~340ms on your typical Rails exception. If you just test something like

```ruby
begin
  raise ArgumentError, "Test"
rescue => ex
  Bugsnag.notify(ex)
end
```

It will be fast because the stack trace is tiny. In your Rails app, it's more common to see 40-60 lines in a stack trace. In your worst case such as IO errors, they can get 3 layers deep of `cause` so you're really cleaning up 160 lines of exceptions which includes an additional ~4 arrays to clean.

It turns out that duping a `Set` is a major time sink. Removing the dups gets it down to 340ms to 40ms. We also call `freeze` on the strings which causes Rails to do some magic and treat them as constants. As well as moving the encoding options to a constant to reduce unnecessary garbage.

https://github.com/zanker/bugsnag_benchmark has an example of the performance of the old code vs the new. I also included a micro optimization one which further improves performance.

This code alone is a ~50% improvement in the benchmark, and in our real world test it's about 90%, as the 340ms -> 40ms change includes the entirety of `Bugsnag.notify`.

### DIFF
--- a/lib/bugsnag/helpers.rb
+++ b/lib/bugsnag/helpers.rb
@@ -4,20 +4,16 @@ module Bugsnag
   module Helpers
     MAX_STRING_LENGTH = 4096
 
-    def self.cleanup_obj(obj, filters = nil, seen=Set.new)
+    def self.cleanup_obj(obj, filters = nil, seen = {})
       return nil unless obj
 
       # Protect against recursion of recursable items
-      if obj.is_a?(Hash) || obj.is_a?(Array) || obj.is_a?(Set)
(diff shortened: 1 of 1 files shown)


ConradIrwin:
Thanks for this. Neat trick with memoizing the clean object in the recursion avoider :).

There's probably plenty of low-hanging performance fruit if you're feeling keen ;)

ConradIrwin:
(released in v2.8.5)

PR #228, opened 2015-04-21
Make sending cookies opt-in

zachmargolis:
For us, cookies often contain fairly sensitive data. In our use case, opting in to cookies makes the most sense but for backwards compatibility I totally understand if you guys want to make cookies opt-out.

cc @nealharris

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -12,6 +12,7 @@ module Bugsnag
     attr_accessor :use_ssl
     attr_accessor :send_environment
     attr_accessor :send_code
+    attr_accessor :send_cookies
     attr_accessor :project_root
     attr_accessor :app_version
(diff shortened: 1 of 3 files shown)


zachmargolis:
Also! I would have added specs, but there don't appear to be any for these middlewares. Happy to write a spec if somebody could point me where it should go.

ConradIrwin:
@zachmargolis I believe if you upgrade to the latest bugsnag-ruby cookies are stripped by default (after https://github.com/bugsnag/bugsnag-ruby/commit/7d40acbcc50a90a951d1e1662d57030fd172a64e). The implementation is a bit subtle, because it adds them in and then stripes them out, but it strips them out of all kinds of places they appear (ENV["HTTP_COOKIE"], env["request.cookie"], etc.)

Please let me know if that's not the case,

Conrad

PR #229, opened 2015-04-29
Report exceptions raised in tasks defined in gems

ConradIrwin:
Before this change gems that were loaded before Bugsnag would
not have the rescue clauses in their tasks.

### DIFF
--- a/lib/bugsnag/rake.rb
+++ b/lib/bugsnag/rake.rb
@@ -2,36 +2,23 @@ require 'bugsnag'
 
 Rake::TaskManager.record_task_metadata = true
 
-module Bugsnag::Rake
-  def self.included(base)
-    base.extend ClassMethods
-    base.class_eval do
-      class << self
-        alias_method :original_define_task, :define_task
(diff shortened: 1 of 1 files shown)


lib/bugsnag/rake.rb
```diff
@@ -2,36 +2,23 @@
...
-        ensure
-          Bugsnag.set_request_data :bugsnag_running_task, old_task
-        end
-      end
-    end
+  rescue Exception => ex
+    Bugsnag.auto_notify(ex)
+    raise
+  ensure
+    Bugsnag.set_request_data :bugsnag_running_task, old_task
```
snmaynard: Why is set_request_data being called here?

ConradIrwin: Rake tasks can call each other so this ensures that exceptions passed to Bugsnag.notify will have the correct task meta-data.

fredrikmollerstrand:
Thanks for this @ConradIrwin!

PR #233, opened 2015-05-15
Add section to README on using with rails-api apps

philiph:
Hi guys.  I was confused as to why the `before_bugsnag_notify` filter was not available within the controllers in my Rails API application.  I've added a section to the README about this, which might save others some time.


snmaynard:
Thanks @philiph!

PR #236, opened 2015-05-29
Mutating @ignore_classes is not threadsafe

mattcross:
Under heavy load on puma we sometimes see ignored exceptions reported to bugsnag.  I believe this is resolved by synchronizing changes to @ignored_classes.

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -31,7 +31,7 @@ module Bugsnag
     attr_accessor :timeout
     attr_accessor :hostname
     attr_accessor :delivery_method
-
+    private_attr_reader :mutex
     attr_writer :ignore_classes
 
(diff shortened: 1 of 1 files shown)


ConradIrwin:
@mattcross Does this actually fix the problem? I think we'd also need to wrap the code that mutates ignore_classes in the same mutex for it to have any effect (and that code is unfortunately in other people's apps).

mattcross:
For rails apps, modifications to ignore_classes should all take place in a rails initializer (or at least during app initialization) so I think the mutex around the map! is all you need.  The mutex definitely works for our case.  You could also have the reader return ```ignore_classes.map { |klass| klass.is_a?(Class) ? klass.name : klass }``` rather than mutating the array and dispense with the mutex.  You would be duplicating a potentially large array many times there but at least it's safe.

ConradIrwin:
Oh thanks, I completely missed that `!`.

PR #239, opened 2015-06-03
Check we respond_to runner

### LINKED ISSUE #238: method_missing: undefined method `runner'
When I add the bugsnag gem to my app I get this error:
```
gems/railties-3.2.13/lib/rails/railtie/configurable.rb:30:in `method_missing': undefined method `runner' for #<Bugsnag::Railtie:0x007f9a7d84b6f0> (NoMethodError)
```
 
Is there some additional configuration that I am missing?

snmaynard:
This could fix #238 but I have yet to test it myself.


snmaynard:
Confirmed that this fixes it on rails 3.2.13.

Interestingly this bug isn't there on 3.2.21 so they must have backported runner in a patch version of rails from 4.

PR #242, opened 2015-06-16
Support faraday error

dtaniwaki:
We could support faraday.

Maybe it should be opt-in?


dtaniwaki:
The notified exception looks like below.

faraday resourcenotfound in kaizen ad bugsnag

jessicard:
Hey @dtaniwaki! Thanks for suggesting this! For now, I don't think we want to *specifically* support Faraday. This can be added using metadata outside of the gem.

PR #244, opened 2015-06-18
Ignore additional exceptions by default

### LINKED ISSUE #230: ActionController::UnknownFormat should be a default ignored class.
This exception is thrown by rails when a request is made with the `Accept` header in a format incorrect for the controller - for example, a html request being made to a json only controller. In production, this is returned as a `406 Not Acceptable` response, and shouldn't show up in bugsnag.

adrianmacneil:
The following exceptions are irrelevant and harmless in production, since they are already handled by rails and will return an appropriate response. They have been causing unnecessary noise for us in Bugsnag.

* `ActionController::ParameterMissing`
* `ActionController::UnknownFormat`
* `ActionController::UnknownHttpMethod`

I also sorted the exception list alphabetically to keep sanity if more ignored exceptions are added in future.

I realize these can be easily configured to be ignored in the bugsnag interface (which we have already done), but I believe they should also be included as default ignore list.

Fixes #230

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -46,15 +46,18 @@ module Bugsnag
     ].freeze
 
     DEFAULT_IGNORE_CLASSES = [
-      "ActiveRecord::RecordNotFound",
-      "ActionController::RoutingError",
-      "ActionController::InvalidAuthenticityToken",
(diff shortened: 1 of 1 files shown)


jessicard:
Thanks @adrianmacneil ! Totally agreed. Thanks for adding them to the list and sorting it :)

PR #245, opened 2015-07-01
Remove multi_json dependency, use 'json' gem for old rubies

loopj:
Takes the exact same approach as `active_support` did when removing `multi_json` here:
https://github.com/rails/rails/pull/10576


loopj:
Takes the exact same approach as `active_support` did when removing `multi_json` here:
https://github.com/rails/rails/pull/10576

PR #246, opened 2015-07-03
Require json instead of multi_json in capistrano.rb

OsQu:
`multi_json` was dropped in v2.8.9, but `capistrano.rb` was still requiring it.

Fix it by requiring `json` instead.


snmaynard:
Thanks @OsQu!

PR #251, opened 2015-07-31
Add back a default timeout of 15 seconds

BlakeMesdag:
At some point the default configuration for timeouts got removed. This adds back a default of 15 seconds which should be plenty for most exceptions.

Our last 24 hours of notifications had a max p95 of 10 seconds for requests, which gives a 5 second runway on open/read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,7 @@ config.proxy_password = "proxy_secret_password_here"
 ```
 
 ###timeout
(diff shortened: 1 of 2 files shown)


snmaynard:
Thanks @BlakeMesdag!

snmaynard:
Released as 2.8.12!

PR #252, opened 2015-08-05
Capistrano - allow for custom endpoint in cap task.

mokhan:
This change allows users of the on-prem version of bugsnag to configure
the custom endpoint to connect to during a capistrano deployment.

The deploy class allows an override of the endpoint attribute [here](https://github.com/bugsnag/bugsnag-ruby/blob/ea2a431f5066ba5c5965da6c3178c778b86c0230/lib/bugsnag/deploy.rb#L11).

Please advise, if I should also bump the version of the gem.


snmaynard:
Thanks @mokhan - I will try and get a release out today!

mokhan:
You're welcome. Thanks for building an excellent product!

snmaynard:
Released as 2.8.12!

PR #253, opened 2015-08-26
Make the list of vendor paths configurable

byroot:
I guess the title explains it all. It allow to mark some part of the application as "framework", that shouldn't be used for grouping, nor title.

@loopj @snmaynard what do you think?

### DIFF
--- a/lib/bugsnag/notification.rb
+++ b/lib/bugsnag/notification.rb
@@ -380,7 +380,7 @@ module Bugsnag
 
         # Generate the stacktrace line hash
         trace_hash = {}
-        trace_hash[:inProject] = true if @configuration.project_root && file.match(/^#{@configuration.project_root}/) && !file.match(/vendor\//)
+        trace_hash[:inProject] = true if in_project?(file)
         trace_hash[:lineNumber] = line_str.to_i
 
         if @configuration.send_code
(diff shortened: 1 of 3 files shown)


lib/bugsnag/notification.rb
```diff
@@ -410,6 +410,17 @@ def stacktrace(exception)
       end.compact
     end
 
+    def in_project?(line)
+      return nil if @configuration.vendor_paths && @configuration.vendor_paths.any? do |vendor_path|
```
snmaynard: I think we should return false here rather than nil to honour the ? convention in ruby.

snmaynard:
I think this is a sensible addition to the library and its a good call to make it backwards compat!

:+1:

lib/bugsnag/notification.rb
byroot: RIght. I'll fix this.

byroot: Done!

snmaynard:
Thanks @byroot!

byroot:
Actually I think I should have included a doc update...

snmaynard:
Thats fine - I was going to throw something in there in a bit myself!

PR #254, opened 2015-09-10
Fix warnings

tfausak:
I encountered some warnings in this gem while reviewing orgsync/stoplight#78. This pull request fixes those warnings. I chose the path of least resistance for most of these fixes. Since I don't know this library, there may be better ways to fix them. 

Ultimately I think the only commit I really care about is 5647d23f5f03f953d789f1be66d0de61ab09caa5. I think it's causing my JRuby build to fail.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
 --backtrace
 --color
 --format doc
+--warnings
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -43,6 +43,7 @@ module Bugsnag
       require "bugsnag/delay/resque" if configuration.delay_with_resque && defined?(Resque)
 
       # Log that we are ready to rock
+      @logged_ready = nil unless defined?(@logged_ready)
       if configuration.api_key && !@logged_ready
         log "Bugsnag exception handler #{VERSION} ready, api_key=#{configuration.api_key}"
         @logged_ready = true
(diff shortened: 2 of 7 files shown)


kattrali:
Hi @tfausak! Thank you for the contribution. I took a look at the crash in JRuby and can confirm the crash and the fix. In addition, shadowing the `config` variable is undesirable and might lead to weird bugs later.

Regarding the use of uninitialized instance variables, this seems to be a fairly standard pattern in Ruby. Have you seen anything in particular which is broken by this syntax? In the meantime, I'm merging this pull request with the caveat of removing 022c3e1 (uninitialized instance variables) and e7aec19 (enable warnings by default). Enabling all warnings turns on a fair amount of chatter from dependencies.

Merged manually in a041d1f.

tfausak:
Using uninitialized instance variables is common in Ruby. I think that is unfortunate, however I don't think it breaks anything. My main motivation for initializing instance variables in this gem was to avoid spurious warning when running the tests in [Stoplight](https://github.com/orgsync/stoplight). The "chatter" you see when enabling warnings is the same chatter that I see in my gem — and I'm trying to reduce it.

kattrali:
> Using uninitialized instance variables is common in Ruby. I think that is unfortunate, however I don't think it breaks anything.

Ah, thank you for the confirmation. 

> My main motivation for initializing instance variables in this gem was to avoid spurious warning when running the tests in Stoplight. The "chatter" you see when enabling warnings is the same chatter that I see in my gem — and I'm trying to reduce it.

Gotcha. I understand not wanting it in your output and share the sentiment. From the bugsnag gem side, enabling warnings picks up a few things in webmock and httpclient which are not ideal but did not seem harmful. I split the dependency warnings into a separate issue to address later (#257). Thank you for following up.

tfausak:
Thanks!

PR #255, opened 2015-09-25
Re-init delivery thread if current PID does not match the one it was created with

jgross206:
Currently, a delivery thread is started one time per process using the
start_once! method.

However, this causes problems in forking code.  If a process encounters
an exception and notifies, then forks, deliveries in the child process
will not be delivered since it will not be given a delivery thread since @started
will be true (inherited from parent process).

This commit modifies the start_once! method to store the PID of the
process in @started.  Then, instead of just checking if @started is set,
it checks if it matched the current PID.  If it doesn't match, the
current process doesn't have a delivery thread so we create one.

### DIFF
--- a/lib/bugsnag/delivery/thread_queue.rb
+++ b/lib/bugsnag/delivery/thread_queue.rb
@@ -26,8 +26,8 @@ module Bugsnag
 
         def start_once!
           MUTEX.synchronize do
-            return if @started
-            @started = true
+            return if @started == Process.pid
+            @started = Process.pid
 
             @queue = Queue.new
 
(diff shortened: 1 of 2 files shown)


jgross206:
shoot, I guess this is failing on jruby because:

1) Bugsnag should work with threadpool delivery after fork
     Failure/Error: Process.fork do
     NotImplementedError:
       fork is not available on this platform

Anyone have any thoughts on what I should do here?

jgross206:
OK, I added a conditional to skip the fork test if the platform is JRuby.  Build passing now

kattrali:
Looks great, thank you @jgross206 for your work on this.

PR #256, opened 2015-10-28
Filter deep parameters, Rails 5-style.

fimmtiu:
Rails 5 has added the ability to filter parameters out of a deeply-nested structure by specifying filters with periods in them, such that (for example) `/users\.id/` will filter `{users: {id: 1}}` to `{users: {id: '[FILTERED]'}}`. I've broken Bugsnag's parameter-filtering code out into a separate class and added similar functionality to make it work with Rails 5-style filters.

For context, see Rails 5's ActionDispatch::Http::ParameterFilter::CompiledFilter:
https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/parameter_filter.rb

Note that I haven't added such functionality to the GET parameter filtering, for two reasons: (1) it would be rather nontrivial to do in full generality, and (2) it only seems to be used in the RackRequest and Rails2Request classes, and neither Rack nor Rails 2 do deep parameter filters.


fimmtiu:
Ping! It's been about six weeks since I created this pull request; is anyone interested in it?

kattrali:
Thank you for the contribution, @fimmtiu. I apologize for the delay. The changes look good. I've had to give some thought to which release to this in, I think it should be grouped with a few other adjustments for rails 5, which should be released together soon.

PR #263, opened 2015-12-04
Reference classes instead of using strings

jgrau:
This fixes the Rails 5 deprecation warning:

```
DEPRECATION WARNING: Passing strings or symbols to the middleware builder is deprecated, please change
them to actual class references.  For example:

  "Bugsnag::Rack" => Bugsnag::Rack
```


kattrali:
This looks good, @jgrau. Thanks!

PR #269, opened 2015-12-28
Test against latest ruby versions

dlackty:


kattrali:
Looks good, thanks @dlackty.

PR #276, opened 2016-01-30
Truncate all event fields if the payload is large

### LINKED ISSUE #247: Some exceptions like JSON::ParserError can generate a very long message which causes Bad Requests 401
The payload_string size is currently checked but only the meta_data hash is reduced. For some exceptions the message can be very long which should be truncated.
In general parsing exceptions generate long messages since they often include the parsed source text.

snmaynard:
This fixes #247


kattrali:
Woot. :+1:

Issue #285, opened 2016-03-23
app_type is not correctly set on sidekiq integration


Draiken:
Hi there, I suspect `Bugsnag::Railtie` class is overriding the `app_type` to `rails` even when the sidekiq integration is loaded. Since the type is set before initialization it means rails is already loaded, so sidekiq's integration had already set the type.

Since sidekiq normally is used alongside rails, I believe this is a bug.

Happy to submit a patch if it is indeed a bug

kattrali:
Thank you for the report, @Draiken! This is definitely a regression and a patch would be welcome.

PR #286, opened 2016-03-29
Ensure Sidekiq app_type is not overwritten when rails is a dependency

Draiken:
This is a fix for #285. It's possible this same bug happens with other integrations, but I only use Sidekiq and didn't verify that.


kattrali:
Thank you for the patch, @Draiken! This is probably a solution which could be applied to Mailman and Resque, though I can take a look at verifying those as well. Looks good.

PR #288, opened 2016-03-30
Truncates stack trace if payload is too long

### LINKED ISSUE #287: Stacktraces can be too long
If a stack overflow exception occurs, the stack trace could be very large, producing a payload that is too large to be processed by Bugsnag.

We could truncate stack traces length if they are too long to ensure that Bugsnag still receives the payload.

CodeHex:
Fixes #287

Limits the number of stack trace frames to 400 if the payload is too large


kattrali:
:tada: Looks good, thanks.

PR #289, opened 2016-04-03
RackRequest - don't allow it to crash on bad URLs and only load Sidekiq middleware if running under Sidekiq server

perldork:
Discovered while using Bugsnag with Rails:
* If the Sidekiq middleware is loaded under Rails, Bugsnag will crash when any error is found during middleware runs, causing the sent over stack trace to not have any custom nor app specific meta-data (for us this meant Rails3, RackRequest, Warden). Use the Sidekiq ```Sidekiq.server?``` test to ensure the middleware is only loaded when running under a Sidekiq process.
* If an improperly formatted URL is sent to a Rails app and it causes a crash, the RackRequest middlleware will crash as well when it attempts to clean the URL's parameters. Add a ```begin/rescue``` statement to keep that from happening so that Rack meta-data still gets sent.

Initially applied to 3.0.0 (version we are running), same patches also applied cleanly to master HEAD.


kattrali:
Thank you for the contribution, @perldork. I merged and squashed this as 5cfae87 externally since there was a previous PR for handling the Sidekiq middleware loading.

perldork:
Delisa,

It was the use statement for Bugsnag::Sidekiq that was causing Bugsnag to
crash, will the previous patch resolve that? i see the use statement is
still i. sidekiq.rb and not wrapped in the CLI conditonal .. my suspicion
that is that if i use that gersion i will get crashes again vs having the
use Bugsnag middleware statement in the CLI conditional in sidekiq.rb

kattrali:
@perldork, the new patch adds a dependency check for Sidekiq, which handles correctly setting app type. This is the same way Bugsnag itself is using the library in production as well as some others. Maybe the crash is unrelated?

perldork:
@kattrali - could be, I will try out the 4.0.0 version in the near future and see what happens - if there are issues i'll see about a patch and open a new PR. thanks for your response.

kattrali:
Sounds good, let me know how it turns out!

aquach:
I'm seeing this behavior where API requests are logged with type sidekiq. To the best of my understanding, bugsnag-ruby loads all the integrations it can find while swallowing load errors, assuming that only the integrations that you've installed will the loaded. But on my Sinatra server, I do need Sidekiq in order to queue Sidekiq jobs. The way I'm reading the code suggests that there's nothing preventing Bugsnag from setting app_type to "sidekiq" on my Sinatra server. Is there something I'm missing here? Thanks for your help!

aquach:
My apologies, I figured it out :)

kattrali:
Ah np, @aquach!

PR #290, opened 2016-04-04
Preserve field types when truncating large payloads

duncanhewett:
When truncating large payloads the event hash is passed into the `reduce_hash_size' helper.  This currently converts booleans and integers to strings which can result the payload being rejected by Bugsnag (particularly the `inProject` field in a stackframe).

This will preserve integers and booleans in the truncated payload.


kattrali:
[Added tests](https://travis-ci.org/bugsnag/bugsnag-ruby/jobs/120734994#L317) and handling for a few edge cases, such as floating-point values and arrays which contains objects other than hashes. Thanks, @duncanhewett!

PR #292, opened 2016-04-17
Add support for 'block syntax' on Bugsnag.notify calls

loopj:
This allows users to customize the error report sent to Bugsnag using ruby block syntax on each `Bugsnag.notify` call as follows:

```ruby
Bugsnag.notify(ex) do |notification|
  notification.severity = "info"
end
```

As well as being an arguably cleaner syntax than before, this also exposes the full power of accessing the `Notification` object to every `Bugsnag.notify` call - provideing a single interface to error report customization. This means `Bugsnag.notify` report customization and `before_notify_callbacks` can all share a single documentation.

This PR is the super un-opinionated version, which does not break existing APIs.

I actually propose we **remove** all the slightly hacky `overrides` logic and replace with this, which would be an api-incompatible change requiring a major version bump - any thoughts?

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -53,15 +53,20 @@ module Bugsnag
     end
 
     # Explicitly notify of an exception
-    def notify(exception, overrides=nil, request_data=nil)
+    def notify(exception, overrides=nil, request_data=nil, &block)
       notification = Notification.new(exception, configuration, overrides, request_data)
+
+      yield(notification) if block_given?
+
       notification.deliver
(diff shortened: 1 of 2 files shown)


snmaynard:
This looks like a good approach to me. I've always felt that notify_or_ignore should be what notify does. If we merge this and bump the major version, how do you feel about making notify and notify_or_ignore do the same thing?

kattrali:
Looks great to me, and seems more Rubyish. Thanks, @loopj!

PR #295, opened 2016-05-11
Trim nested arrays and drop metadata to save on payload size

kattrali:
When a payload is too large, the request is rejected by the Bugsnag API. This changeset adds a few more last resort adjustments to error reports to ensure at least something is sent.

If a payload is too large to be sent, the following steps are taken, and then only if the previous step did not put the payload size under the threshold:

1. **Trim strings.** All strings over 3kb are truncated. This still leaves enough room for the opening paragraphs of [War and Peace](http://www.gutenberg.org/files/2600/2600-h/2600-h.htm#link2HCH0001).
2. **Truncate arrays.** All arrays are truncated to having 40 items or less. If an array of 40 items has maximum length strings, ~5kb remains for the remainder of the payload
3. **Drop metadata.** Custom metadata attached to the report is dropped. This should ensure that a report is at least generated, even if there is less available to assist with debugging.

### DIFF
--- a/lib/bugsnag/helpers.rb
+++ b/lib/bugsnag/helpers.rb
@@ -5,7 +5,7 @@ require 'json' unless defined?(JSON)
 
 module Bugsnag
   module Helpers
-    MAX_STRING_LENGTH = 4096
+    MAX_STRING_LENGTH = 3072
     MAX_PAYLOAD_LENGTH = 128000
     MAX_ARRAY_LENGTH = 40
     RAW_DATA_TYPES = [Numeric, TrueClass, FalseClass]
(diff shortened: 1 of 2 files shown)


lib/bugsnag/helpers.rb
```diff
@@ -101,12 +101,27 @@ def self.truncate_arrays_in_value(value)
...
       when Array, Set
-        truncate_arrays(value)
+        truncate_array(value)
       else
         value
       end
     end
 
+    def self.remove_metadata_from_value(object)
+      case object
```
snmaynard: This would be clearer and faster if it were more explicit.

```ruby
payload.events.map do |event|
  event[:metaData] = {}
end
```

kattrali: Makes sense.

PR #296, opened 2016-05-23
Update README.md instructions for capistrano integration

drakmail:
All capistrano plugins must be required in `Capfile` – updated that in README


kattrali:
Thank you for the contribution @drakmail! I've been working to move the reference documentation into a new hosted site, so I merged your changeset there. [Here is the new deploy tracking guide](http://docs.bugsnag.com/platforms/ruby/rails/deploy-tracking/#using-capistrano).

PR #299, opened 2016-06-07
Don't log API key

jborrey:
[This PR](https://github.com/bugsnag/bugsnag-ruby/commit/5eaddd862ea6b0c39ac9ecef070a8a94befd6e56) was a good start but there was more.

Used to find where more changes were needed
```bash
~/src/bugsnag-ruby (master)$ grep -r "#{api_key}" .
./lib/bugsnag/notification.rb:        Bugsnag.warn "Your API key (#{api_key}) is not valid, couldn't notify"
./lib/bugsnag/notification.rb:        Bugsnag.log("Notifying #{endpoint} of #{@exceptions.last.class} from api_key #{api_key}")
```

Should not log API key since logs are not necessarily kept as secret a
environment variables. The log steam should be keeping track of which
service/box is doing the logging in order to identify which client of
the API is having issues.

Result of running spec:
```
Finished in 3.78 seconds (files took 0.86747 seconds to load)
130 examples, 0 failures
```

### DIFF
--- a/lib/bugsnag/notification.rb
+++ b/lib/bugsnag/notification.rb
@@ -219,7 +219,7 @@ module Bugsnag
 
         # Build the endpoint url
         endpoint = (@configuration.use_ssl ? "https://" : "http://") + @configuration.endpoint
-        Bugsnag.log("Notifying #{endpoint} of #{@exceptions.last.class} from api_key #{api_key}")
+        Bugsnag.log("Notifying #{endpoint} of #{@exceptions.last.class}")
 
         # Deliver the payload
         self.class.deliver_exception_payload(endpoint, build_exception_payload, @configuration, @delivery_method)


kattrali:
Thank you for the contribution, @jborrey. Taking a look.

lib/bugsnag/notification.rb
```diff
@@ -181,7 +181,7 @@ def deliver
         Bugsnag.warn "No API key configured, couldn't notify"
         return
       elsif api_key !~ API_KEY_REGEX
-        Bugsnag.warn "Your API key (#{api_key}) is not valid, couldn't notify"
```
kattrali: This instance should probably stay; since the API key does not match the required format, its useful to compare the logged output to your actual key to correct typos. There is no data exposure for logging an incorrect key.

jborrey: Do you think that it is still a dangerous log line, say if someone accidentally set the Bugsnag API key to another API key or malformed API key?
Would the message not give enough information without logging the API key?

If not, I'll not remove this log line but would prefer too.

kattrali: It's not dangerous at all to keep it logged, really. The API key cannot be used to access your data in any way. It does seem excessive however that it is logged everywhere always.

jborrey:
@kattrali Updated so it only removes on instance of the API key logging.

kattrali:
Thanks! Releasing this in a patch shortly.

PR #301, opened 2016-06-16
Update doc with current list of default ignore_classes

tjwp:
Also, update default `params_filters`.


kattrali:
Thank you for the contribution, @tjwp! The documentation is moving over to a new hosted site for the reference content, so I pulled in your changeset there. [The updated content is available here](http://docs.bugsnag.com/platforms/ruby/rails/configuration-options/#ignore_classes). I'll follow with a commit removing the old content in `docs/`.

PR #305, opened 2016-06-29
Que integration

s-andringa:
Integrates [Que](https://github.com/chanks/que). Also deals with ActiveJob-scheduled Que jobs.

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -130,7 +130,7 @@ module Bugsnag
   end
 end
 
-[:resque, :sidekiq, :mailman, :delayed_job].each do |integration|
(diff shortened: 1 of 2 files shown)


kattrali:
Whoa, this is cool. I apologize for the delay, looking at this now.

s-andringa:
Thanks @kattrali. One thing to note: if a custom Que error handler is set after Bugsnag-ruby is initialized, then the error handler that Bugsnag sets will be overruled by it. One could overcome this by calling the original handler from the new handler, e.g. by doing something similar to this:
```rb
original_handler = Que.error_handler
Que.error_handler = proc do |error, job|
  
  # Custom handling here...

  original_handler.call(error, job) if original_handler
end

PR #308, opened 2016-07-21
Remove RoutingError from default ignore classes

snmaynard:
ActionController::RoutingError is raised when calling a router function that doesnt exist and can generate a 500 page. I think not having this in defaults is safer


rickmzp:
Rails's `ActionDispatch::ExceptionWrapper`'s `:not_found` responses include both `ActionNotFound` and `RoutingError`, which means that Rails wants to treat these as 404s.

We just upgraded to Bugnsag 5 and are now getting exceptions reported for every single 404 and we're not entirely sure how to handle it now that this is removed.

`ActionNotFound` is actually more likely to be an exceptional case than `RoutingError`, which can be triggered by a user simply mistyping a URL.

Issue #298, opened 2016-05-25
Prefer BUGSNAG_RELEASE_STAGE over RAILS_ENV for release_stage


dziemba:
We deploy our projects to multiple release stages, but they all have `RAILS_ENV` set to `production`. I think this is a good way to keep the environments in sync, [Heroku also agree with me](https://devcenter.heroku.com/articles/deploying-to-a-custom-rails-environment).

When using the bugsnag gem, it currently overrides `BUGSNAG_RELEASE_STAGE` set in the environment with the current `RAILS_ENV` value. I think it would be better that `RAILS_ENV` is only used if `BUGSNAG_RELEASE_STAGE` is not set.

I have a workaround for this, but I'd prefer not having to add this to all our projects.

```ruby
Bugsnag.configure do |c|
  c.release_stage = ENV.fetch('BUGSNAG_RELEASE_STAGE') if ENV.key?('BUGSNAG_RELEASE_STAGE')
end
```

What do you think about changing the precedence of `RAILS_ENV` and `BUGSNAG_RELEASE_STAGE`?

kattrali:
I think that makes sense, thank you for the suggestion, @dziemba.

snmaynard:
This has now been addressed on master and I will do a release soon!

dziemba:
Thank you very much for this!

timdiggins:
@snmaynard this is really useful thanks -- could you add it to bugsnag official docs for ease of reference? 

e.g. https://docs.bugsnag.com/platforms/ruby/rails/#automatically-captured-diagnostics
![bugsnag_docs_ _platforms_ _ruby_ _rails](https://cloud.githubusercontent.com/assets/18395/22513691/e4c090dc-e894-11e6-9414-d6cb88fad1ce.png)

(add something like  `will use environment variable BUGSNAG_RELEASE_STAGE or RAILS_ENV`)

PR #310, opened 2016-08-10
Give precedence to the BUGSNAG_RELEASE_STAGE environment variable

s-andringa:
Implements #298


GrahamCampbell:
This is technically a breaking change, right?

s-andringa:
Yes, if people have `BUGSNAG_RELEASE_STAGE` set to something different than the Rails/Rack env the behaviour changes.

kattrali:
Thank you for the contribution, @s-andringa. Changing this override behavior makes sense, as any Rails/Rack user who attempting to override `BUGSNAG_RELEASE_STAGE` had no effect. This change is only breaking in the sense that now the environment variable would do something.

snmaynard:
This seems good to me too - merging down now

snmaynard:
Thanks @s-andringa!

PR #314, opened 2016-08-17
Don't override Rails 5 belongs_to association behavior

mateuszgorniak:
Hi!
I've found a problem with default behavior of `belongs_to` association.
In Rails 5 ["belongs_to" association is required by default](https://github.com/rails/rails/pull/18937).

Example scenario:

```
class CustomModel < ApplicationRecord
  belongs_to :association
  # validates :association, presence: true is not necessary anymore
end
```

After adding your gem I've noticed that this change is not working anymore:

```
m = CustomModel.new
m.valid? #=> true but should return false
m.errors.full_messages #=> [] but should include proper error message
```

My PR should fix this problem.

### DIFF
--- a/lib/bugsnag/railtie.rb
+++ b/lib/bugsnag/railtie.rb
@@ -47,8 +47,11 @@ module Bugsnag
         ActionController::API.send(:include, Bugsnag::Rails::ControllerMethods)
       end
       if defined?(ActiveRecord::Base)
-        require "bugsnag/rails/active_record_rescue"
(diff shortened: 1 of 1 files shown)


rafaelfranca:
This can be closed now.

emaiax:
So, this is going to be merged anytime soon?

kattrali:
@emaiax @mateuszgorniak Sorry about that, this change was brought in as a part of a separate changeset, and this PR wasn't closed appropriately. This change is in v5.1.0.

emaiax:
@kattrali awesome! thanks! :)

Issue #309, opened 2016-08-10
Better error context for Sidekiq


sandstrom:
This commit adds the class: https://github.com/bugsnag/bugsnag-ruby/commit/9d79c9f4698dcafbb28d5c1af139b97d8a6d058d

However, it's just *ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper* in Bugsnag (not very useful):

![image](https://cloud.githubusercontent.com/assets/122287/17546349/f717eede-5ee1-11e6-930b-75a0b6e3326c.png)

Instead I'd suggest using the `wrapper` value (which is shown in the Sidekiq tab). This will be something like `MyQuickJob` which is much more useful.

kattrali:
Hi @sandstrom, thanks for the feedback. That's a good idea. I was thinking that class name was something more useful, but perhaps the wrapper makes more sense.

sandstrom:
@kattrali The class name never change, for all our errors it always says `ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper` instead of the ActiveJob class.

snmaynard:
Just got a fix for this in #317 - will get a release out soon

sandstrom:
@snmaynard Awesome! ⛵  When will you release a new version?

snmaynard:
Planning on a release early next week - probably Monday or Tuesday. A few more things I want to try and get in to it.

sandstrom:
@snmaynard much thanks

sandstrom:
@snmaynard We've deployed 5.0 and still seeing this. Have I misunderstood something or is the issue still around?

*(expected the context value, to the right of the bold error class to be the name of the job)*

![image](https://cloud.githubusercontent.com/assets/122287/18045387/e396df00-6dd2-11e6-9ab3-dc12a673b33c.png)

sandstrom:
@snmaynard friendly ping 😄

snmaynard:
#323 should be the fix for this - we should be using `wrapped` not `wrapper` for the active job class name

kattrali:
Closed via #323.

Issue #306, opened 2016-07-07
Bugsnag.clear_request_data not called after Sidekiq worker runs for Sidekiq versions >= 3


tmak:
Maybe I'm missing something here.
For Sidekiq versions >= `3` the middleware [`Bugsnag::Sidekiq`](https://github.com/bugsnag/bugsnag-ruby/blob/master/lib/bugsnag/sidekiq.rb#L4) is not used and because of that [`Bugsnag.clear_request_data`](https://github.com/bugsnag/bugsnag-ruby/blob/master/lib/bugsnag/sidekiq.rb#L17) is not called after Sidekiq worker runs.

I'm happy to create a PR if somebody else confirms this as a bug.

martin308:
Hi @tmak, as we don't call `set_request_data` in the code path when sidekiq version is >= 3 there is no need to call `clear_request_data` as the middleware does.

snmaynard:
I have just merged in a fix for this in #317 - will get a release out for it soon

PR #317, opened 2016-08-19
Sidekiq improvements

### LINKED ISSUE #282: Set resque/sidekiq errors to synchronous notify
If we detect we are running in sidekiq or resque, can we synchronously notify to ensure delivery?

snmaynard:
Fixes #282, #309, #306

### DIFF
--- a/lib/bugsnag/railtie.rb
+++ b/lib/bugsnag/railtie.rb
@@ -7,8 +7,6 @@ require "bugsnag/middleware/rack_request"
 
 module Bugsnag
   class Railtie < Rails::Railtie
-    cattr_accessor :running_as_dependency
-
     rake_tasks do
       require "bugsnag/rake"
       load "bugsnag/tasks/bugsnag.rake"
@@ -51,7 +49,7 @@ module Bugsnag
         ActiveRecord::Base.send(:include, Bugsnag::Rails::ActiveRecordRescue)
       end
 
-      Bugsnag.configuration.app_type = "rails" unless Bugsnag::Railtie.running_as_dependency
+      Bugsnag.configuration.app_type = "rails"
     end
 
     # Configure params_filters after initialization, so that rails initializers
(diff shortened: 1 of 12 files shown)


lib/bugsnag/railtie.rb
```diff
@@ -51,7 +49,7 @@ class Railtie < Rails::Railtie
         ActiveRecord::Base.send(:include, Bugsnag::Rails::ActiveRecordRescue)
       end
 
-      Bugsnag.configuration.app_type = "rails" unless Bugsnag::Railtie.running_as_dependency
+      Bugsnag.configuration.app_type = "rails"
```
snmaynard: This change relies on the fact that in the other cases (sidekiq etc) we know that we are running in that env. This is definitely a little broad, but I think its OK. This will ultimately be set to rails in the console and in the server case, which is what we want.

It is also initially set to rails in other apps (for example in a sidekiq app that also has rails) but because we later know we are running in a sidekiq app, we change the app_type then to be sidekiq.

snmaynard:
I think this also fixes some bugs with app_type detection and adds some example apps for testing/demo purposes.

kattrali:
LGTM

lib/bugsnag/resque.rb
```diff
@@ -37,4 +37,6 @@ def save
 # Auto-load the failure backend
 Bugsnag::Resque.add_failure_backend
 
-Bugsnag.configuration.app_type = "resque"
+Resque.before_first_fork do
+  Bugsnag.configuration.app_type = "resque"
```
snmaynard: Here I should also add synchronous delivery mechanism

lib/bugsnag/sidekiq.rb
```diff
@@ -2,9 +2,14 @@
 
 module Bugsnag
   class Sidekiq
+    def initialize(options = {})
```
martin308: is there a reason for the options param?

snmaynard: None any more - removed.

lib/bugsnag/resque.rb
```diff
@@ -37,4 +37,7 @@ def save
...
+  Bugsnag.configuration.delivery_method = :synchronous
```
rafaelfranca: Why is this setting synchronous? How does we override this value in resque workers if we don't want it?

PR #319, opened 2016-08-20
Alias notify_or_ignore to notify

snmaynard:
`notify` and `notify_or_ignore` causes confusion with getting ignore classes to work.

I propose that we release this as is as it lines up with the documentation (we dont actually document `notify_or_ignore`) and then in a larger v5 rewrite I have going we clean this up and break the interface.

Thoughts?

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -58,16 +58,6 @@ module Bugsnag
 
       yield(notification) if block_given?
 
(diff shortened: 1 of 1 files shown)


kattrali:
Makes sense to me. We could also warn via the logger that `notify_or_ignore` is deprecated and then call through to `notify`.

PR #320, opened 2016-08-21
V6

snmaynard:
This is a fairly big rewrite/change for the ruby notifier.

Its not complete but I wanted to deprecate lots of old interfaces for the ruby notifier.
## Changes

-`config.ignore_classes` now no longer accepts strings
- `Bugsnag.notify_or_ignore` removed - now call `notify` directly
- `Bugsnag.auto_notify` removed - now call `notify` directly
- `config.delay_with_resque` removed
- `Bugsnag::Notification` now `Bugsnag::Report`
- `notify` now only supports the block, you can no longer send in overrides hash
- `config.use_ssl` removed, now set protocol in `config.endpoint`
- `config.vendor_paths` removed (Should we leave this in for people that locally install gemfile?)
- `after_notify_callbacks` removed
- Capistrano support removed - will go into a dedicated gem
- Deploy support removed, will go in the capistrano gem
- `params_filters` renamed to `meta_data_filters` to be clearer
- `config.debug` removed and instead use the logger directly

Probably more besides. This cannot be merged until the capistrano gem exists.


lib/bugsnag/tasks/bugsnag.rake
```diff
@@ -30,55 +9,6 @@ namespace :bugsnag do
...
-
-      # Add the hook
-      url = "https://notify.bugsnag.com/deploy?" + params.map {|k,v| "#{k}=#{v}"}.join("&")
-      command = "heroku addons:add deployhooks:http --url=\"#{url}\""
-      command += " --app #{ENV["HEROKU_APP"]}" if ENV["HEROKU_APP"]
-
-      puts "$ #{command}"
-      run_command.call(command)
-    end
-  end
```
kattrali: Was this part supposed to go into bugsnag-capistrano? Seems odd now to have it in either place since the deployment stuff moved there.

snmaynard: We could move it to bugsnag-api?

kattrali:
Closing for #385 (resolves conflicts).

PR #321, opened 2016-08-29
Support ActiveJob wrapped delayed jobs

abraham-chan:
screen shot 2016-08-29 at 4 04 32 pm

Better support for ActiveJob wrapped delayed jobs. The payload details like `job_class` and `arguments` are now shown in the "Job" tab in bugsnag


jonathan-wheeler:
Is there anything that needs doing for this to get merged?

PR #327, opened 2016-10-06
Support for configurable exclusion patters in stack traces

blanquer:
Adds support for configurable exclusion patterns in reported stacktrace lines.
- Added new configuration key: `stacktrace_exclusion_patterns`. This key will contain an array of Regex patterns which when matched, will exclude said lines from the reported the stack trace in the exceptions.
  - By default the exclusion pattern will contain the original regexp (`%r{lib/bugsnag(/|\.rb)}`
  - Usage example: `Bugsnag.configuration.stacktrace_exclusion_patterns << /mypat/`

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -33,6 +33,7 @@ module Bugsnag
     attr_accessor :hostname
     attr_accessor :delivery_method
     attr_writer :ignore_classes
+    attr_accessor :stacktrace_exclusion_patterns
 
     THREAD_LOCAL_NAME = "bugsnag_req_data"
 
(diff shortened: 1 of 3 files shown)


Cawllec:
Hey, this is very easy to setup using callback blocks now.
To filter particular files out of the stacktrace you can do:
```
Bugsnag.notify(RuntimeError.new("oops")) do |report|
    exception[:stacktrace].reject! do |line|
      /stack_trace_line.rb/.match(line[:file])
    end
  end
end
```

If you're not doing this on a case-by-case basis and want to attach the callback to all notifications then use:
```
Bugsnag.before_notify_callbacks << lambda {|report|
    exception[:stacktrace].reject! do |line|
      /stack_trace_line.rb/.match(line[:file])
    end
  end
}
```

PR #329, opened 2016-10-19
Add more to delayed job notifications

timdiggins:
Currently the Bugnsag DelayedJob Plugin / DelayedJob notifier doesn't show ActiveJob data. This PR addresses that.

It also adds the queue (if present) for the DJ job and the #attempts

runtimeerror_in_fnc_backend_webapp

PS: Hard to work out how to get any specs for this to run (I did try...), as DelayedJob isn't actually a dependency for this gem (and there are no existing specs for the DJ plugin code), but if you can give some guidance on this bit of set up I could write them.


timdiggins:
Failing only on ruby 2.2.4 -- see #330

PR #331, opened 2016-11-09
Example Chef integration - first draft

gkze:
- [x] Report failed Chef runs to Bugsnag

https://docs.chef.io/handlers.html#event-types

**To do**
- [ ] Add more examples

CC: @kattrali: I still have to figure out good conventions for how other Chef events should be reported

### DIFF
--- a/example/chef/Gemfile
+++ b/example/chef/Gemfile
@@ -1,0 +1,4 @@
+source 'https://rubygems.org'
+
+gem 'bugsnag'
+gem 'chef'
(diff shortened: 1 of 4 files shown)


kattrali:
> I still have to figure out good conventions for how other Chef events should be reported

Well, it makes sense to report with the information that would make it easier to debug. Maybe that means the Chef environment, certain variables, or adding custom context. The block argument of `notify` might be useful here, until we know what we should aggregate automatically.

gkze:
Agreed. I'll add more information in the notify call about the failure.

PR #334, opened 2016-11-17
Use File.exist? instead of File.exists?

Manfred:
In newer versions of Ruby using File.exists? will warn about deprecation. The File.exist? method has been around forever so this will not introduce any backwards compatibility problems.


kattrali:
Thanks, @Manfred! That makes sense. :)

PR #335, opened 2016-11-17
[Resque] Do not leak config in parent process

kirs:
Mutating `Bugsnag.configuration` in `before_first_fork` means that it will also affect the parent process. But in this case, we only want to mutate configuration for the new forked Resque process.

@keeganlow

cc @casperisfine @fw42

### DIFF
--- a/lib/bugsnag/resque.rb
+++ b/lib/bugsnag/resque.rb
@@ -37,7 +37,7 @@ Resque::Failure::Bugsnag = Bugsnag::Resque
 # Auto-load the failure backend
 Bugsnag::Resque.add_failure_backend
 
-Resque.before_first_fork do
+Resque.after_fork do
   Bugsnag.configuration.app_type = "resque"
   Bugsnag.configuration.delivery_method = :synchronous
 end


fw42:
What if the app doesn't use Resque in forking mode but runs all jobs inline?

fw42:
Also, what if there is an exception in the parent? Shouldn't those be delivered synchronously too?

kirs:
> Also, what if there is an exception in the parent? Shouldn't those be delivered synchronously too?

It will be delivered with the parent delivery method. I think it's fine.

> What if the app doesn't use Resque in forking mode but runs all jobs inline?

Then `before_first_fork` won't be executed anyway, doesn't it?

casperisfine:
> What if the app doesn't use Resque in forking mode but runs all jobs inline?

Is that even a thing?

> what if there is an exception in the parent?

Bugsnag hookpoint into Resque make it so that only errors happening inside the job would be reported. An exception in Resque itself would let Resque crash.

fw42:
> Then before_first_fork won't be executed anyway, doesn't it?

I think that hook always runs, even if you never work.

> Is that even a thing?

yes, that's what we do in Shopify. We don't fork per job.

kattrali:
Thank you for the contribution, @kirs. Taking a look. 👀 

> Mutating `Bugsnag.configuration` in `before_first_fork` means that it will also affect the parent process. But in this case, we only want to mutate configuration for the new forked Resque process.

Why is this necessary? Would you mind sharing a bit more about your Resque setup?

casperisfine:
> Would you mind sharing a bit more about your Resque setup?

Right, it's true that it's absolutely not obvious.

In our main app we extend Resque to add a few concerns, and we test those extensions in our test suite. Some of those tests boot a Resque worker, which trigger this callback.

As a result the Bugsnag configuration is mutated for the entire test suite which makes other tests fail later. This is what we call leaky tests internally.

We have other ways to fix this, mostly by restoring the Bugsnag configuration to it's original value after each test, but if it wasn't mutated in the first place it would be even better for us.

But if this change doesn't work for all ways you could configure Resque, we're fine with not merging this change.

lib/bugsnag/resque.rb
```diff
@@ -37,7 +37,7 @@ def save
 # Auto-load the failure backend
 Bugsnag::Resque.add_failure_backend
 
-Resque.before_first_fork do
+Resque.after_fork do
   Bugsnag.configuration.app_type = "resque"
   Bugsnag.configuration.delivery_method = :synchronous
```
rafaelfranca: Setting this by default means that users can not choose a different delivery method when loading this code. I have comment this same problem here https://github.com/bugsnag/bugsnag-ruby/pull/317#pullrequestreview-9967661.

Is it possible to not change the delivery_method, or at least check if the user already configured a different value before overriding it?

snmaynard: Why would you want to make a change? Just trying to understand the use case. Certain values of this will prevent you being notified so I just want to be careful with this.

rafaelfranca: In Shopify we have a custom delivery method that we use for all the types of workers (web and jobs). We are aware that using async delivery method may not work for jobs, but not being possible for us to use our custom delivery method in jobs is blocking us to upgrade this gem.

snmaynard: Well that doesnt sound good, we should try to get a solution for this. Its tricky because detection around if we are running in a resque app happens later than the configure block is called - so we will have to make sure we get the precedence right here of those options.

snmaynard: I've raise an internal ticket to work on this in the next couple of weeks. You should hear about the issue discussed here and the broader resque config soon.

kirs: I think we should avoid magic here (automatically detecting Resque and setting the config option) and let Rescue/Sidekiq users to add declare configuration setting themselves. This way it will be explicit and it won't hiddenly break someone's app by modifying global state from a `fork` hook.

martin308:
@kirs @fw42 @casperisfine I have made a suggestion as to how we can alter this to satisfy everything raised here, check it out in #347 and let me know what you think

casperisfine:
Can't comment on #347 because of some GitHub bug, but :+1:

Thanks @martin308!

kattrali:
Merged via #347!

PR #337, opened 2016-11-26
prevent db connections during asset precompile

karlhungus:
#### Problem

We are creating db connections during asset compilation

#### Solution

Do connections on load

#### Problem

We are creating db connections during asset compilation

#### Solution

Do connections on load

Full disclosure: i have no idea what i'm doing :(

@kirs @rafaelfranca /cc @celsodantas

### DIFF
--- a/lib/bugsnag/railtie.rb
+++ b/lib/bugsnag/railtie.rb
@@ -45,8 +45,10 @@ module Bugsnag
         ActionController::API.send(:include, Bugsnag::Rails::ControllerMethods)
       end
       if defined?(ActiveRecord::Base)
-        require "bugsnag/rails/active_record_rescue"
-        ActiveRecord::Base.send(:include, Bugsnag::Rails::ActiveRecordRescue)
+        ActiveSupport.on_load(:active_record) do
+          require "bugsnag/rails/active_record_rescue"
+          ActiveRecord::Base.send(:include, Bugsnag::Rails::ActiveRecordRescue)
+        end
       end
 
       Bugsnag.configuration.app_type = "rails"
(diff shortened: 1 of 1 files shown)


lib/bugsnag/railtie.rb
```diff
@@ -45,8 +45,10 @@ class Railtie < Rails::Railtie
         ActionController::API.send(:include, Bugsnag::Rails::ControllerMethods)
       end
       if defined?(ActiveRecord::Base)
-        require "bugsnag/rails/active_record_rescue"
-        ActiveRecord::Base.send(:include, Bugsnag::Rails::ActiveRecordRescue)
+        ActiveSupport.on_load(:active_record) do
+          require "bugsnag/rails/active_record_rescue"
+          ActiveRecord::Base.send(:include, Bugsnag::Rails::ActiveRecordRescue)
```
rafaelfranca: `include Bugsnag::Rails::ActiveRecordRescue`

lib/bugsnag/railtie.rb
```diff
@@ -45,8 +45,10 @@ class Railtie < Rails::Railtie
         ActionController::API.send(:include, Bugsnag::Rails::ControllerMethods)
       end
       if defined?(ActiveRecord::Base)
```
rafaelfranca: This is not needed anymore.

rafaelfranca:
This is also related with https://github.com/bugsnag/bugsnag-ruby/pull/314

kattrali:
This makes sense. :) Thank you for the contribution, @karlhungus, and for looking it over, @rafaelfranca.

PR #338, opened 2016-11-28
Correctly hook on Action Controller

rafaelfranca:
When an initializer load the `ActionController::Base` constant all the on_load triggers for `:action_controller` are executed. This cause a lot of load order issues.

Gems should always use `ActiveSupport.on_load` to extent behavior of the Rails frameworks.

Also using `ActiveSupport.on_load(:action_controller)` we can make sure that both `ActionController::Base` and `ActionController::Api` are extended.

This is related with #337 and #314.

### DIFF
--- a/lib/bugsnag/railtie.rb
+++ b/lib/bugsnag/railtie.rb
@@ -37,12 +37,9 @@ module Bugsnag
       config = YAML.load_file(config_file) if File.exist?(config_file)
       Bugsnag.configure(config[::Rails.env] ? config[::Rails.env] : config) if config
 
-      if defined?(::ActionController::Base)
+      ActiveSupport.on_load(:action_controller) do
         require "bugsnag/rails/controller_methods"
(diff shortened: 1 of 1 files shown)


emaiax:
I hope this can be merged soon. :shipit:

cupakromer:
**ping** 💙 

Hoping this gets merged soon. Just ran into a strange issue with Rails 5 `ActionController::API` and `render :action` (I am using Jbuilder for the renderer). All requests return the string body of `" "`. Applying this patch resolves the issue (as does removal of the `bugsnag` gem).

This can easily be reproduced with a new Rails 5.0.1 app, Jbuilder 2.6.1, and bugsnag 5.0.1. Here is the complete controller and view I used:

```ruby
# app/controllers/widgets_controller.rb
class WidgetsController < ActionController::API
  def index
    render :index
  end
end
```

```ruby
# app/views/widgets/index.json.jbuilder
json.data "API Testing"
```

kattrali:
Thank you! This will be in a release shortly.

PR #350, opened 2017-03-17
Prepend Bugsnag in the Sidekiq middleware chain

akhiln:
Starting in version 3.3.0, Sidekiq introduced the ability to prepend to the middleware chain. We should prepend when we can so that we can catch any exceptions that occur in the middleware chain.


kattrali:
Looks good. Thanks @akhiln!

PR #355, opened 2017-04-14
Pass temporary queue to check resque worker fork_per_job?

### LINKED ISSUE #354: Resque.workers.new is complaining because called without queues
This commit breaks if I update 5.2.0 -> 5.3.0.

https://github.com/bugsnag/bugsnag-ruby/commit/091704afc425e50b400538a339e7222343d34f9e

I reverted this and I am good for now. Should be an easy fix :)

FreekingDean:
Quick fix to use a queue name when initializing a worker.
Use the queue name `:bugsnag_fork_check` for the initialized worker that is used to call `.fork_per_job?`
fixes #354


kattrali:
Works perfectly! Thanks, @FreekingDean.

PR #359, opened 2017-04-28
Add support for DelayedJob Custom Job arguments

calvinhughes:
This adds support for passing the arguments for [Custom Job objects](https://github.com/collectiveidea/delayed_job#custom-jobs) in the DelayedJob integration.

Please advise of the best way to handle writing any specs for this. I've tested it manually on various job types.

### DIFF
--- a/lib/bugsnag/delayed_job.rb
+++ b/lib/bugsnag/delayed_job.rb
@@ -33,7 +33,12 @@ unless defined? Delayed::Plugins::Bugsnag
               p[:id]           = payload.id           if payload.respond_to?(:id)
               p[:display_name] = payload.display_name if payload.respond_to?(:display_name)
               p[:method_name]  = payload.method_name  if payload.respond_to?(:method_name)
-              p[:args]         = payload.respond_to?(:args) ? payload.args : payload.to_h
+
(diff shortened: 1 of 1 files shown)


lib/bugsnag/delayed_job.rb
```diff
@@ -33,7 +33,8 @@ def error(job, error)
               p[:id]           = payload.id           if payload.respond_to?(:id)
               p[:display_name] = payload.display_name if payload.respond_to?(:display_name)
               p[:method_name]  = payload.method_name  if payload.respond_to?(:method_name)
-              p[:args]         = payload.args         if payload.respond_to?(:args)
+              p[:args]         = payload.respond_to?(:args) ? payload.args : payload.to_h
```
kattrali: Where does `to_h` come from? I didn't see this mentioned in DelayedJob anywhere.

calvinhughes: @kattrali It's available as a method on the `Struct` class which is passed to `Delayed::Job#enqueue`. This outputs the parameters to that job in the form of a hash.

However I can see it might raise problems where the user has a Custom class that does not support this method, so I'll update it to not add `args` in such case. What do you think?

PR #361, opened 2017-06-13
Rescue exceptions raised when calling to_s while sanitizing

kattrali:
In the case that calling to_s will raise, like in `ActionDispatch::RemoteIp::IpSpoofAttackError`, rescue and insert a placeholder.

[Relevant example](https://github.com/rails/rails/blob/34fe2a4fc778d18b7fe6bdf3629c1481bee789b9/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L134)


snmaynard:
Does this mean that every to_s will display [SPOOF] if it throws? Any reason we cant to_s earlier the ip in the middleware earlier and rescue [SPOOF] there?

kattrali:
We can.

PR #364, opened 2017-08-24
Modified test and travis system to check dependency separation

Cawllec:
Looking to update both the travis.yml and Gemfile to ensure support for libraries such as sidekiq and coverage tools do not cause issues in testing/ci


kattrali:
The rabbit hole for what was going on with CI was way deeper than I first thought. I made a few small commits to deal with bundler weirdness and then behavior mismatches between tests and Ruby versions.

PR #368, opened 2017-09-13
Added unhandled functionality, updated tests and integrations

Cawllec:
- Adds `defaultSeverity`, `unhandled`, and `severityReasons` payload properties to support upcoming handled/unhandled functionality
- Created separate notifier functions, `notify` for handled notifications and `auto_notify` for automated/middleware notifications
- Middleware set up with correct severity reasons
- Discussion needed around language-level ruby error handling, and how to allow the user to this up without delving into severity reasons
_references [PLAT_207](https://bugsnag.atlassian.net/browse/PLAT-207)_

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -33,36 +33,13 @@ module Bugsnag
     end
 
     # Explicitly notify of an exception
-    def notify(exception, &block)
+    def notify(exception, auto_notify=false, &block)
       report = Report.new(exception, configuration)
-      send_report(report, false, &block)
-    end
 
-    # For automatic notification of exceptions
-    def auto_notify(exception, severity_reason, &block)
-      if !configuration.auto_notify
+      if !configuration.auto_notify && auto_notify
         configuration.debug("Not notifying because auto_notify is disabled")
         return
       end
-
-      report = Report.new(exception, configuration, true, severity_reason)
-
-      send_report(report, true, &block)
-    end
-
-    # Configuration getters
-    def configuration
-      @configuration = nil unless defined?(@configuration)
-      @configuration || LOCK.synchronize { @configuration ||= Bugsnag::Configuration.new }
-    end
-
-    # Allow access to "before notify" callbacks
-    def before_notify_callbacks
-      Bugsnag.configuration.request_data[:before_callbacks] ||= []
-    end
-
-    private
-    def send_report(report, auto_notify, &block)
 
       if !configuration.valid_api_key?
         configuration.debug("Not notifying due to an invalid api_key")
@@ -107,9 +84,7 @@ module Bugsnag
         end
 
         # Test whether severity has been changed
-        if report.severity != initial_severity
-          report.default_severity = false
-        end
+        report.default_severity = report.severity == initial_severity
 
         # Deliver
         configuration.info("Notifying #{configuration.endpoint} of #{report.exceptions.last[:errorClass]}")
@@ -117,6 +92,17 @@ module Bugsnag
         configuration.debug("Payload: #{payload_string}")
         Bugsnag::Delivery[configuration.delivery_method].deliver(configuration.endpoint, payload_string, configuration)
       end
+    end
+
+    # Configuration getters
+    def configuration
+      @configuration = nil unless defined?(@configuration)
+      @configuration || LOCK.synchronize { @configuration ||= Bugsnag::Configuration.new }
+    end
+
+    # Allow access to "before notify" callbacks
+    def before_notify_callbacks
+      Bugsnag.configuration.request_data[:before_callbacks] ||= []
     end
   end
 end
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -33,11 +33,36 @@ module Bugsnag
     end
 
     # Explicitly notify of an exception
-    def notify(exception, auto_notify=false, &block)
-      if auto_notify && !configuration.auto_notify
+    def notify(exception, &block)
+      report = Report.new(exception, configuration)
+      send_report(report, false, &block)
+    end
+
+    # For automatic notification of exceptions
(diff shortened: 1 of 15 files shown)


lib/bugsnag/integrations/resque.rb
```diff
@@ -26,7 +26,12 @@ def self.add_failure_backend
     end
 
     def save
-      Bugsnag.notify(exception, true) do |report|
+      Bugsnag.auto_notify(exception, {
+        :type => "middleware",
```
snmaynard: middleware_handler

lib/bugsnag.rb
```diff
@@ -33,12 +33,37 @@ def configure
...
+    end
+
+    # For automatic notification of exceptions
+    def auto_notify(exception, severity_reason, &block)
+      if !configuration.auto_notify
         configuration.debug("Not notifying because auto_notify is disabled")
         return
       end
 
+      report = Report.new(exception, configuration, true, severity_reason)
```
snmaynard: You could set the severity reason in here to be error?

lib/bugsnag/tasks/bugsnag.rake
```diff
@@ -6,7 +6,14 @@ namespace :bugsnag do
     begin
       raise RuntimeError.new("Bugsnag test exception")
     rescue => e
-      Bugsnag.notify(e, {:context => "rake#test_exception"})
+      Bugsnag.auto_notify(e, {
```
snmaynard: This is fine to just be a normal notify, doesnt need to be an auto_notify. Its just a way of people testing their bugsnag installations

lib/bugsnag.rb
```diff
@@ -33,12 +33,37 @@ def configure
...
     # Explicitly notify of an exception
-    def notify(exception, auto_notify=false, &block)
-      if auto_notify && !configuration.auto_notify
+    def notify(exception, &block)
+      report = Report.new(exception, configuration)
+      send_report(report, false, &block)
+    end
+
+    # For automatic notification of exceptions
+    def auto_notify(exception, severity_reason, &block)
```
snmaynard: I dont think we need the auto_notify method. If you look where we build and send the report blocks when auto_notify is set to true are run first so we can change the reports in the block for auto notifications.

lib/bugsnag.rb
```diff
@@ -80,24 +106,18 @@ def notify(exception, auto_notify=false, &block)
           return
         end
 
+        # Test whether severity has been changed
+        if report.severity != initial_severity
+          report.default_severity = false
```
snmaynard: You could make this a little clearer imo with `report.default_severity = report.severity == initial_severity`

lib/bugsnag.rb
snmaynard: So I would just call notify like we did before, just telling notify that this is an auto_notify. Then in the block you can set the severity reason and the severity. Its weird that severity is set in the block, and severity reason outside it right now.

Cawllec:
I agree with re-merging the auto-notify functions, and with passing the severity_reason through a block, as passing it along with the severity makes the most sense.

lib/bugsnag.rb
```diff
@@ -34,7 +34,9 @@ def configure
 
     # Explicitly notify of an exception
     def notify(exception, auto_notify=false, &block)
-      if auto_notify && !configuration.auto_notify
+      report = Report.new(exception, configuration, auto_notify)
```
kattrali: Why create the report before the validation blocks? This is extra processing when a user is running in a testing/debug mode.

lib/bugsnag.rb
```diff
@@ -65,13 +65,27 @@ def notify(exception, auto_notify=false, &block)
...
       configuration.middleware.run(report) do
         if report.ignore?
           configuration.debug("Not notifying #{report.exceptions.last[:errorClass]} due to ignore being signified in user provided middleware")
           return
         end
 
+        # Update severity reason if necessary
+        if report.severity != before_middleware_severity
+          report.severity_reason = {
+            :type => "userSpecifiedSeverity"
```
kattrali: The values for severity_reason should be defined as constants to avoid typos and to aggregate all possible values. Currently these values are scattered between different files.

lib/bugsnag/integrations/rack.rb
```diff
@@ -35,6 +35,12 @@ def call(env)
         # Notify bugsnag of rack exceptions
         Bugsnag.notify(raised, true) do |report|
           report.severity = "error"
+          report.set_handled_state({
+            :type => "unhandledExceptionMiddleware",
+            :attirbutes => {
```
kattrali: "attirbutes"

lib/bugsnag/integrations/rack.rb
```diff
@@ -45,6 +51,12 @@ def call(env)
       if env["rack.exception"]
         Bugsnag.notify(env["rack.exception"], true) do |report|
           report.severity = "error"
+          report.set_handled_state({
+            :type => "middleware_handler",
+            :attributes => {
+              :name => "rack"
```
kattrali: `Rack` above is capitalized, while this is not. Integration names are another candidate for a constant enumeration.

lib/bugsnag/middleware/classify_error.rb
```diff
@@ -0,0 +1,45 @@
...
+        "SystemExit"
+    ]
+
+    def initialize(bugsnag)
+      @bugsnag = bugsnag
+    end
+
+    def call(report)
+      report.raw_exceptions.each do |ex|
+        ancestor_chain = ex.class.ancestors.select { |ancestor| ancestor.is_a?(Class) }.map { |ancestor| ancestor.to_s }
```
kattrali: Breaking the blocks onto separate lines would make this expression more readable.

PR #372, opened 2017-09-25
Added 'did you mean' suggestions middleware

Cawllec:
- Adds middleware to capture 'did you mean' suggestions and append them to meta-data in ruby version > 2.3.0
- Uses string matching to not break earlier versions
- Includes tests to ensure early ruby version run correctly

### DIFF
--- a/lib/bugsnag/middleware/suggestion_data.rb
+++ b/lib/bugsnag/middleware/suggestion_data.rb
@@ -2,7 +2,7 @@ module Bugsnag::Middleware
   class SuggestionData
 
     CAPTURE_REGEX = /Did you mean\?([\s\S]+)$/
-    DELIMITTER = "\n"
+    DELIMITER = "\n"
 
     def initialize(bugsnag)
       @bugsnag = bugsnag
@@ -14,12 +14,12 @@ module Bugsnag::Middleware
         match = CAPTURE_REGEX.match(exception.message)
         next unless match
         
-        suggestions = match.captures[0].split(DELIMITTER).each { |suggestion|
+        suggestions = match.captures[0].split(DELIMITER).each { |suggestion|
           matches[matches.size] = suggestion.strip
         }
       end
 
-      report.add_tab(:"did you mean", matches) if matches.size > 0
+      report.add_tab(:suggestion, matches) if matches.size > 0
       
       @bugsnag.call(report)
     end
--- a/lib/bugsnag/middleware/suggestion_data.rb
+++ b/lib/bugsnag/middleware/suggestion_data.rb
@@ -1,0 +1,27 @@
+module Bugsnag::Middleware
+  class SuggestionData
+
+    CAPTURE_REGEX = /Did you mean\?([\s\S]+)$/
+    DELIMITTER = "\n"
+
+    def initialize(bugsnag)
+      @bugsnag = bugsnag
+    end
+
+    def call(report)
+      matches = {}
+      report.raw_exceptions.each do |exception| 
+        match = CAPTURE_REGEX.match(exception.message)
+        next unless match
+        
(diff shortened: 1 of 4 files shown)


lib/bugsnag/middleware/suggestion_data.rb
```diff
@@ -0,0 +1,27 @@
+module Bugsnag::Middleware
+  class SuggestionData
+
+    CAPTURE_REGEX = /Did you mean\?([\s\S]+)$/
+    DELIMITTER = "\n"
```
kattrali: I think the extra 'T' in "delimiter" is a typo here

lib/bugsnag/middleware/suggestion_data.rb
```diff
@@ -0,0 +1,27 @@
...
+      report.raw_exceptions.each do |exception| 
+        match = CAPTURE_REGEX.match(exception.message)
+        next unless match
+        
+        suggestions = match.captures[0].split(DELIMITTER).each { |suggestion|
+          matches[matches.size] = suggestion.strip
+        }
+      end
+
+      report.add_tab(:"did you mean", matches) if matches.size > 0
```
kattrali: Is there a case for this to happen multiple times in a single exception? Otherwise I think we can better name the metadata field something like `suggestion` so that it is filterable by users rather than using an index.

Cawllec: I've changed the tab name to `suggestions`, is there a better method of attaching the suggestions than indexed in a hash?  It seems like the library blocks directly attaching an array as a tab.

kattrali: Naming the tab "did you mean" is fine, the matches could instead be placed in a hash:

```ruby
{suggestions: matches}
```

Though since there's usually one, we should instead unwrap it:

```ruby
if matches.size == 1
  report.add_tab(:"did you mean", {suggestion: matches.first}
elsif matches.size > 1
  # add suggestions array, as above
end
```

lib/bugsnag/middleware/suggestion_data.rb
```diff
@@ -0,0 +1,30 @@
...
+      report.raw_exceptions.each do |exception| 
+        match = CAPTURE_REGEX.match(exception.message)
+        next unless match
+        
+        suggestions = match.captures[0].split(DELIMITER)
+        matches.concat suggestions.map{ |suggestion| suggestion.strip }
+      end
+
+      if matches.size == 1
+        report.add_tab(:"did you mean", {:suggestion => matches.first})
```
snmaynard: Commonly in other notifiers we use an error tab i believe to generically have information about the error? Having a tab with one entry in it doesnt seem like something we would want to do long term

kattrali: That's a good idea. We could use "Error" as the tab name, in case there is other error-specific items to add later.

snmaynard: As an idea I just had - its where the bugsnag_meta_data attached to the error could go?

Cawllec: Would that risk making the 'custom' tab redundant?  It might be useful creating an 'error' tab and renaming the 'custom' tab to separate error generated vs user generated

kattrali: @Cawllec The custom tab is where metadata is collated if it does not conform to the expected format (e.g., not nested so that there is a tab name and a key name, like if you use `report.add_tab(:error, "foo")`). Its a bit out of scope of this changeset, but I think I see what you mean. Let's use `:error` as the tab name here though, and then we could separately use it as the tab for `bugsnag_meta_data` in a later feature.

PR #373, opened 2017-09-26
Adding a breadcrumb logger 

Cawllec:
## Additions
- Adds a logger to track events output to a logger to breadcrumbs
- `BugsnagLogger` can be assigned to any interface that uses a Logger, such as `ActiveRecord::Base.logger` or `Mongoid::Loggable.logger`
- Breadcrumb logging threshold can be set during initialisation or at any point afterwards
- `MultiLogger` accepts an array of loggers on creation and will output to each of them on message receipt, allowing user to attach other loggers alongside `BugsnagLogger`

## Discussion
- Considering changing name to `BreadcrumbLogger` to be more explicit about functionality
- I created this as a separate PR to [the breadcrumbs PR](https://github.com/bugsnag/bugsnag-ruby/pull/367) to avoid splitting focus.  I think the breadcrumb feature should be completed before this

## TODO
- Needs tests added
- Needs examples/documenting

### DIFF
--- a/lib/bugsnag/loggers/bugsnag_logger.rb
+++ b/lib/bugsnag/loggers/bugsnag_logger.rb
@@ -1,0 +1,110 @@
+require "bugsnag"
+require "bugsnag/breadcrumbs/breadcrumb"
+
+module Bugsnag::Loggers
+
+  LEVELS = [
+    "debug",
+    "info",
+    "warn",
+    "error",
+    "fatal"
+  ]
+
+  class BugsnagLogger
+
+    attr_reader :level
+
+    def initialize(level="info")
+      @level = level
+    end
+
+    def add(severity, message = nil, progname = nil)
+      if supports_level?(severity)
+        Bugsnag.leave_breadcrumb(message, Bugsnag::Breadcrumbs::LOG_TYPE, {
+          :progname => progname,
+          :severity => severity
+        })
+      end
+    end
+
+    def <<(message)
+      add "unknown", message
+    end
+
+    alias :sev_threshold :level
+    def level(severity)
+      if Loggers::LEVELS.include? severity
+        @level = severity
+      end
+    end
+
+    def info(message)
+      add "info", message
+    end
+
+    def info(progname=nil, &block)
+      yield message="" if block_given?
+      add "info", message, progname
+    end
+
+    def info?
+      supports_level?("info")
+    end
+
+    def debug(progname=nil, &block)
+      yield message="" if block_given?
+      add "debug", message, progname
+    end
+
+    def debug?
+      supports_level?("debug")
+    end
+
+    def error(progname=nil, &block)
+      yield message="" if block_given?
+      add "error", message, progname
+    end
+
+    def error
+      supports_level?("error")
+    end
+
+    def fatal(progname=nil, &block)
+      yield message="" if block_given?
+      add "fatal", message, progname
+    end
+
+    def fatal?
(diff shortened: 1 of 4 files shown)


Cawllec:
Tests are now added as well as a few updates

lib/bugsnag/loggers/multi_logger.rb
```diff
@@ -0,0 +1,85 @@
+module Bugsnag::Loggers
```
snmaynard: I dont think we should implement this. We can recommend another one if we want, but its out of the scope of our lib imo

lib/bugsnag/loggers/bugsnag_logger.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  LEVELS = [
+    "debug",
+    "info",
+    "warn",
+    "error",
+    "fatal"
+  ]
+
+  class BugsnagLogger
```
snmaynard: we can call this logger and put it in the bugsnag module if we get rid of multi

lib/bugsnag/loggers/bugsnag_logger.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    attr_reader :level
+
+    def initialize(level="info")
```
snmaynard: We havent inherited from logger which is potentially a better idea. I looked at some things. Found log4r which might be a good thing to use? We could also instead build an appender? http://www.rubydoc.info/gems/logging/Logging/Appenders 

That way we dont have to even worry about multi loggers. I havent looked into this deeply at all however!

Cawllec: I'll subclass Appender/IO for the same goal in a different branch.  If we'd prefer that I'll close this one out

Cawllec: Hmm, I didn't notice when I last checked that Appenders doesn't belong to Stdlib.  Personally I'd prefer to implement an stdlib solution with examples for other common third parties.  I can remove multi-logger and re-tool Bugsnag-Logger into just Logger.

kattrali: There's two sets of users here that we can support:

* folks using the stdlib logger
* folks using other loggers (e.g. logging and log4r gems)

For the first group, the simplest solution seems to be subclassing [`Logger`](http://ruby-doc.org/stdlib-2.4.2/libdoc/logger/rdoc/Logger.html), since it already provides log levels and filtering by level. In that case the implementation only needs to map levels to severities and make breacrumbs/reports from the `log()` method when its invoked.

For the folks using something other than the stdlib Logger, the [logging](https://github.com/TwP/logging) gem seems to be the most used and maintained, as log4r hasn't seen a release in the last 7ish years. We'd need an appender for this and a check that the `Logging` class exists to activate it. The appender and the stdlib Logger subclass could share the same logic, potentially, depending on what an appender needs to do.

Cawllec: stdlib includes a [`LogDevice`](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger/LogDevice.html) which seems to be equivalent to an Appender in the Logging lib.  I think targeting these two and allow users to configure the actual logger may be the best solution.

Cawllec:
I think we're in favour of #375 more than this, so I'll close this for now.

PR #375, opened 2017-10-03
Added basic logger/log-device and tests

Cawllec:
## Additions
- Adds a logger to track events output to a logger to breadcrumbs
- `Logger` can be assigned to any interface that uses a Logger, such as `ActiveRecord::Base.logger` or `Mongoid::Loggable.logger`
- Adds an appender for use with the [logging library](https://github.com/TwP/logging)
- `Appender` includes trackable `LogEvent` data including `file`, `line`, `method`


lib/bugsnag/loggers/logger.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def add(severity, message = nil, progname = nil)
+      return unless @open
+      if block_given?
+        message = yield message
+      elsif message.nil?
+        message = progname
+      end
+      if severity >= level
+        @log_device.write(message, progname, Bugsnag::Loggers::SEVERITIES[severity])
```
kattrali: Is it possible to define your own `level`? What happens if its over `Logger::UNKNOWN`?

lib/bugsnag/loggers/logger.rb
```diff
@@ -0,0 +1,53 @@
...
+      return unless @open
+      @log_device.write message
+    end
+
+    def close
+      @open = false
+      true
+    end
+
+    def reopen
```
kattrali: What will this be used for?

lib/bugsnag/loggers/logger.rb
```diff
@@ -0,0 +1,53 @@
...
+    "fatal",
+    "unknown"
+  ]
+
+  class Logger < Logger
+
+    def initialize(level=Logger::INFO)
+      @open = true
+      @log_device = Bugsnag::Loggers::LogDevice.new
+      super @log_device, level
```
kattrali: What is the benefit of having the device over `nil` here? The `add()` method could call `Bugsnag.leave_breadcrumb` directly.

lib/bugsnag/loggers/logger.rb
Cawllec: On the base class it's meant to close the connection to a file or IO object.  While we're not hanging onto those objects I thought it'd be appropriate to cease logging in similar circumstances

lib/bugsnag/loggers/logger.rb
Cawllec: You're right, there is no advantage.  I'll remove this.  For some reason I assumed the logger base class wouldn't be happy being passed nil, but it's quite happy.

lib/bugsnag/loggers/logger.rb
```diff
@@ -0,0 +1,68 @@
+require "logger"
+require "bugsnag"
+
+module Bugsnag::Loggers
```
snmaynard: Is there any benefit to the Loggers module? `Bugsnag::Logger` reads nicer. Is there a clash with our internal logger maybe?

Cawllec: There's going to be some shared functionality between the `Logger` and the `Appender` classes (Appender coming soon).  I've thought about renaming the module 'logging' as it's what the module deals with, but it will contain 'loggers', so not sure about that

Cawllec:
Added the appender in with tests, I'll get some examples added, one using `Appender` in a general purpose way and one using `Logger` to record breadcrumbs from activeRecord

Cawllec:
This is becoming part of a separate project

PR #376, opened 2017-10-06
Fix NameError when exception is raised in delayed_job

eitoball:
This PR should fix the following exception is raised when exception is raised in delayed_job.

`NameError: uninitialized constant Notification::UNHANDLED_EXCEPTION_MIDDLEWARE`


Cawllec:
Thanks for the PR eitoball, I've tweaked and merged in #377

Issue #371, opened 2017-09-19
Bugsnag client forces Net::HTTP to ignore http_proxy env variable


dexhorthy:
Howdy -- wanna start by thanking the whole bugsnag team -- we are big bugsnag fans over here, keep up the great work! We've found one issue (this behavior may even be intentional) around handling of http proxies. Would be happy to work on a patch for this.

### Expected behavior
Busnag respects `http_proxy` env var when set, and doesn't pass `nil` to `Net::HTTP` for the proxy address.

```ruby
    uri = URI.parse("https://some-destination.example.com")
    ENV['http_proxy'] = "http://some-proxy:3128"

    http = Net::HTTP.new(uri.host, uri.port) 
    request = Net::HTTP::Get.new(uri.request_uri)
    response = http.request(request) # request will be routed through proxy from ENV
```

### Observed behavior
Bugsnag always passes `nil` for the http proxy address unless it is explicitly set via `config.proxy_host`. This overrides the default `Net::HTTP` behavior of transparently using the proxy set in `ENV['http_proxy']`. This happens here: 

https://github.com/bugsnag/bugsnag-ruby/blob/42552a894bb2f70fd48df8b6bc9d3e3da7ba11aa/lib/bugsnag/delivery/synchronous.rb#L27

```ruby
    uri = URI.parse("https://some-destination.example.com")
    ENV['http_proxy'] = "http://some-proxy:3128"

    # if config.proxy_host, port, etc is set to nil, those will be passed in here
    http = Net::HTTP.new(uri.host, uri.port, nil, nil, nil, nil) # nil passed for proxy, wont use proxy from env
    request = Net::HTTP::Get.new(uri.request_uri)
    response = http.request(request) # not routed through proxy
```

### Steps to reproduce
Set `http_proxy` env var, configure bugsnag without modifying proxy configuration. Requests to bugsnag API will be not be routed through proxy.

### Version

`  bugsnag (~> 2.0)`

### Additional information

Notably, `Net::HTTP` doesn't transparently support proxies with authentication, so there are still many cases where users of this client would need to manually configure Bugsnag's http proxy settings. That should always be the source of truth if set, but I'm wondering if y'all would be open to allowing configuration via the `http_proxy` env var, and not passing in `nil` for the proxy address if those configuration options are unset.

This could technically be a breaking change if anyone was relying on the bugsnag client to explicity pass `nil` for the proxy address.

[...]

PR #379, opened 2017-10-06
Enabled setting proxy_host through environment variables

Cawllec:
- Enables setting of proxy_host from Environment variable as described [here](https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-Proxies)
- Fixes the behaviour described [here](https://github.com/bugsnag/bugsnag-ruby/issues/371)


kattrali:
This change will also require a line in `UPGRADING.md` indicating that the default behavior for proxy handling now reads from the environment.

kattrali:
lgtm

PR #380, opened 2017-10-10
Fixes #371 allow transparent proxy config in `Net::HTTP` via ENV

### LINKED ISSUE #371: Bugsnag client forces Net::HTTP to ignore http_proxy env variable

dexhorthy:


Cawllec:
Thanks @dexhorthy, I'll create a similar PR for our upcoming version to ensure this isn't left out.

dexhorthy:
Sounds good -- thanks for having a look!

PR #393, opened 2017-11-14
Use lazy load hooks to hook on ActionController::Base:

Edouard-chin:
👋  Hello guys thanks for creating and maintaining this gem

Use lazy load hooks to hook on ActionController::Base:

- Calling directly ActionController::Base affects the initialization process (one of them being one that controls ActionController configuration)
- As a result setting configuration inside an initializer (such as the asset_host), doesn't have any effect

### DIFF
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -43,17 +43,14 @@ module Bugsnag
         config.middleware.insert_before Bugsnag::Middleware::Callbacks, Bugsnag::Middleware::Rails3Request
       end
 
-      if defined?(::ActionController::Base)
+      ActiveSupport.on_load(:action_controller) do
         require "bugsnag/integrations/rails/controller_methods"
-        ::ActionController::Base.send(:include, Bugsnag::Rails::ControllerMethods)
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Someone reverted my change https://github.com/bugsnag/bugsnag-ruby/pull/338 😢

lib/bugsnag/integrations/railtie.rb
```diff
@@ -43,17 +43,17 @@ class Railtie < Rails::Railtie
...
       end
 
-      if defined?(::ActionController::Base)
-        require "bugsnag/integrations/rails/controller_methods"
-        ::ActionController::Base.send(:include, Bugsnag::Rails::ControllerMethods)
-      end
-      if defined?(ActionController::API)
-        require "bugsnag/integrations/rails/controller_methods"
-        ActionController::API.send(:include, Bugsnag::Rails::ControllerMethods)
+      if defined?(ActionController::Base) || defined?(ActionController::API)
```
rafaelfranca: You don't need to check. Just define the hooks.

kattrali:
Thanks for this, @Edouard-chin @rafaelfranca. We're going to take a look and figure out where a merge went awry.

PR #395, opened 2017-11-16
Mark file paths beginning with 'vendor' as not in project

### LINKED ISSUE #394: Bring back vendor paths
Before v6.0, Bugsnag was annotating each line of the backtrace, setting `inProject` to `false` if the file was not in the project root, or if it was in one of the vendor paths, which defaulted to directories named `vendor`.

I think this feature should be brought back, because when you use `bundle install --deployment`, by default it installs the gems inside the project, under `vendor/bundle`. 

### Expected behavior
* I expect to only see my code when I look at an error or event on Bugsnag, not the framework's code or gems.
* If I need to, I can show the full trace.

### Observed behavior
* All lines of the backtrace end up showing on the dashboard.
* It seems the grouping algorithm uses the visible lines, so now unrelated events can get grouped together if they share the first line of the stacktrace inside a gem (typically for me it's an HTTP client like Faraday), or even from the stdlib like Net::HTTP (e.g. on Heroku ruby is installed in the project root, under `vendor/`, see [heroku-buildpack-ruby](https://github.com/heroku/heroku-buildpack-ruby/blob/49af4ace3f7517adf0544cfd24d21759948c8e3e/lib/language_pack/ruby.rb#L181)).

### Version
* v6.0.0 and later

kattrali:
Fixes #394

### DIFF
--- a/lib/bugsnag/stacktrace.rb
+++ b/lib/bugsnag/stacktrace.rb
@@ -7,6 +7,9 @@ module Bugsnag
     # e.g. "org.jruby.Ruby.runScript(Ruby.java:807)"
     JAVA_BACKTRACE_REGEX = /^(.*)\((.*)(?::([0-9]+))?\)$/
 
+    # Path to vendored code. Used to mark file paths as out of project.
+    VENDOR_PATH = 'vendor/'
+
     def initialize(backtrace, configuration)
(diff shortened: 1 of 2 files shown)


Cawllec:
I think this can probably be released as a patch with #396

PR #397, opened 2017-11-20
Add at_exit block automatically

Cawllec:
Adds a configuration options and method to register a pre-defined at_exit block when configuration is run.

- `register_at_exit` function accessible on library to add at_exit block with severity reasoning included
- `add_exit_handler` configuration option automatically calls `register_at_exit` if set to true, and is flagged to only do so once
- `add_exit_handler` defaults to false to avoid issues in frameworks (for instance can block `exit 1` signals in rails)


lib/bugsnag.rb
```diff
@@ -114,6 +120,21 @@ def notify(exception, auto_notify=false, &block)
       end
     end
 
+    # Registers an at_exit function to automatically catch errors on exit
+    # This can disable by setting the 'add_exit_handler' option to False
```
snmaynard: This can be disabled by setting the 'add_exit_handler' option to false

Cawllec:
I put it on a separate config option as at_exit blocks can interfere with exit codes.  I didn't think it was a good idea to have at_exit blocks being registered automatically if users just want the rails/other framework exception hooks working.

I don't believe it captures the kill signal, but I'm not sure it won't spit out the last exception in the app, even if it's been captured and dealt with already.  I'll test this tomorrow.

snmaynard:
I'm pretty sure it will capture kill

snmaynard:
It doesnt look like the code here will alter the exit code?

Cawllec:
From my testing it doesn't capture kill signals in the example rails apps.  Nor does it re-send the previous error if it's already been handled.  Rather than linking it to auto_notify I'd rather have it opt-out, but still separate.

snmaynard:
Why? Python links it to auto_notify. Why the difference?

Cawllec:
The main difference is that this change could potentially be considered "breaking" for users who're already using custom at_exit blocks which we've advised users to do in the documentation to this point.  
If there's already an at_exit error handler in place it will be sending two error notifications per exit than one.

snmaynard:
How does it sound to put it behind a separate flag for now but when we do a major version bump we make a change to make it consistent with python? I like consistency across the notifiers because it makes support and integration easier for users.

Cawllec:
Provided @duncanhewett is happy with the documentation I'd like to get this merged and released.

snmaynard:
I would prefer to bump to v7 and make the at_exit block gated by auto_notify and on by default.

Cawllec:
I'd be happy with that, I'll have a look at what other changes could be made at the same time.

Cawllec:
Tests appear to be failing on this due to travis issues with Jruby. It installs and runs through tests fine locally.

lib/bugsnag.rb
```diff
@@ -126,6 +124,40 @@ def notify(exception, auto_notify=false, &block)
       end
     end
 
+    ##
+    # Registers an at_exit function to automatically catch errors on exit
+    # This can be disabled by setting the 'add_exit_handler' configuration option to false
```
kattrali: The comment is outdated now

Cawllec:
Can I get a re-review on this @snmaynard?

snmaynard:
I pushed a cleanup commit to this branch addressing my review comments.

Cawllec:
Thanks for that.  For CI: Rubocop is complaining about module length - I would suggest again bumping up the module limit size and bringing forward a planned refactor.

PR #400, opened 2017-11-21
Add initializers to rails examples

Cawllec:
Adds initialisers to rails examples, ensuring they're runnable with minimal setup.


kellymason:
Gemfile.lock for each example should be checked in, yeah?

PR #401, opened 2017-11-21
Rails 4.2 example: remind users to run the migration

kellymason:
Should we advise users that, to properly run this example, they'll need to run the CreateDelayedJobs migration?


snmaynard:
Is it not the same for all examples?

Cawllec:
I believe this is the only one with delayed jobs integrated into it.  Did I miss that from the readme?

Issue #345, opened 2017-02-03
Proposal: add ability to silence exception instances


johnnyshields:
I have some cases where I do:

```ruby
rescue MyException => e
  Bugsnag.notify(e, {meta: data})
  raise e
end
```

The exception is raised twice in Bugsnag. I'd prefer if I could do something like:

```ruby
rescue MyException => e
  Bugsnag.notify(e, {meta: data})
  e.bugsnag_silence!
  raise e
end
```

And that instance wouldn't generate a Bugsnag when it hit's Bugsnag's global catching logic.

PR #402, opened 2017-11-24
Added support for skip_bugsnag flag on exceptions

Cawllec:
Adds supports for setting a `skip_bugsnag` flag on exceptions to allow re-raising after notifying as requested in #345

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -62,6 +62,11 @@ module Bugsnag
         return
       end
 
+      if exception.respond_to?(:skip_bugsnag) && exception.skip_bugsnag
+        configuration.debug("Not notifying due to skip_bugsnag flag")
+        return
+      end
+
       report = Report.new(exception, configuration, auto_notify)
 
       # If this is an auto_notify we yield the block before the any middleware is run
--- a/spec/report_spec.rb
+++ b/spec/report_spec.rb
@@ -1000,6 +1000,14 @@ describe Bugsnag::Report do
     }
   end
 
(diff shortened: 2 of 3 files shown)


snmaynard:
Why add this rather than advise the use of `bugsnag_meta_data`? This will report the exception as handled, when infact it is unhandled.

Cawllec:
For the particular use case of the guy in #345 just using the `bugsnag_meta_data` wouldn't necessarily solve his problem, whereas this allows specific individual exceptions to be skipped if required.

snmaynard:
It would solve his problem - he could attach the metadata and then reraise

javierjulio:
@snmaynard when you refer to `bugsnag_meta_data`, do you mean I should do the following to avoid double notification if I end up re-raising the error?

```ruby
rescue SomeStandardErrorClass => error
  Bugsnag.notify(error, { skip_bugsnag: true })
  raise error
end
```

The Bugsnag docs state setting a `skip_bugsnag` property on an error class which I realize now doesn't make any sense since where I tend to do this type of handling is for third party code (e.g. API client library) where I can't set that. The method is not defined.

Would the metadata approach work in the example I gave? Should [the docs example](https://docs.bugsnag.com/platforms/ruby/rails/#avoiding-re-notifying-exceptions) be updated?

snmaynard:
I think you would want to do something like this,

```ruby
rescue SomeStandardErrorClass => error
  Bugsnag.notify(error)
  error.define_singleton_method(:skip_bugsnag) {true}
  raise error
end
```

Issue #383, opened 2017-10-12
Successfully exiting from a rake task creates a bugsnag error


leesmith:
### Expected behavior
No bugsnag error created.

### Observed behavior
Bugsnag error created:

```
Unhandled error:  SystemExit: exit
```

### Steps to reproduce

Create a rake task that successfully exits with a success exit code:

```
exit 0
```

### Version
Bugsnag 5.4.1
Rails 5.1.4

### Additional info
I didn't see this behavior using Bugsnag 5.3.3.

Thanks for any help!

csuhta:
Having a similar issue, we have a rake task that runs as a worker process.

Part of the code looks like this:

```ruby
Signal.trap("SIGTERM") { 
  # [...]
  exit(0) 
}
```

Bugsnag traps the `SystemExit` as a problem. My Bugsnag error ID is `59e4829828b3d200197587a5` if you want to inspect your record for it.

Cawllec:
Thanks for the reports, I'll have a look into it.

Cawllec:
So due to issues with certain exceptions not being reported and a lack of good reasoning behind ignoring certain error classes by default we've removed the default `ignore_classes` list.

The previously ignored classes are now instead marked as info severity but will still be reported.

If you want to add these classes back in to your project you can configure them [as described in the docs](https://docs.bugsnag.com/platforms/ruby/configuration-options/#ignore_classes).

kattrali:
Seems pretty clear here that `SystemExit` is a meaningful default value and we should have it in `ignore_classes` by default. And, knowing why, its easy to document.

csuhta:
I would also suggest restoring `Interrupt`  to the default list of ignores too. (The exception raised when someone CTRL-C’s the process)

PR #404, opened 2017-12-05
Add SystemExit & Interrupt to ignore_classes by default

Cawllec:
Re-adds commonly seen exit 'exceptions' to default ignore_classes set to prevent them automatically triggering a notification.
- Adds `SystemExit` and `Interrupt` to `ignore_classes` set by default.
- Fixes issue described in #383


leesmith:
Thank you!

PR #410, opened 2018-01-02
Testing coverage improvement

Cawllec:
An effort to ensure the tests cover more of the notifier.
Adds:
  - Middleware tests
  - Integrations tests
  - Warden/Clearance user tests
  - Improves existing test coverage

This will be ongoing until all integrations are covered by basic tests.


Cawllec:
I think this is probably in a good place to think about merging in now, and further updates can be done on a continuous basis.

PR #412, opened 2018-01-09
Session tracking fixes

Cawllec:
Fixes several issue from initial session-tracking release including:
- Standardises API
- Reduces number of spawned threads
- Uses concurrent ruby for safer concurrency
- Fixed breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Changelog
 ### Enhancements
 
 * Adds support for tracking sessions and crash rate by setting the configuration option `configuration.track_sessions` to `true`.
-  Sessions can be manually created using `Bugsnag.start_session`, and manually delivered using `Bugsnag.send_sessions`.
+  Sessions can be manually created using `Bugsnag.create_session`, and manually delivered using `Bugsnag.send_sessions`.
   | [#411](https://github.com/bugsnag/bugsnag-ruby/pull/411)
 
 ## 6.4.0 (21 Dec 2017)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@ Changelog
 
 ### Enhancements
 
-* Adds support for tracking sessions and crash rate by setting the configuration option `configuration.track_sessions` to `true`.
-  Sessions can be manually created using `Bugsnag.create_session`, and manually delivered using `Bugsnag.send_sessions`.
+* Adds support for tracking sessions and crash rate by setting the configuration option `configuration.auto_capture_sessions` to `true`.
+  Sessions can be manually created using `Bugsnag.start_session`.
   | [#411](https://github.com/bugsnag/bugsnag-ruby/pull/411)
 
 ## 6.4.0 (21 Dec 2017)
--- a/bugsnag.gemspec
+++ b/bugsnag.gemspec
@@ -18,5 +18,5 @@ Gem::Specification.new do |s|
   ]
   s.require_paths = ["lib"]
   s.required_ruby_version = '>= 1.9.2'
-  s.add_runtime_dependency 'concurrent-ruby', '~> 1.0.5'
+  s.add_runtime_dependency 'concurrent-ruby', '~> 1.0'
 end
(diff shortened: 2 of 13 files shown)


bugsnag.gemspec
```diff
@@ -18,4 +18,5 @@ Gem::Specification.new do |s|
   ]
   s.require_paths = ["lib"]
   s.required_ruby_version = '>= 1.9.2'
+  s.add_runtime_dependency 'concurrent-ruby', '~> 1.0.5'
```
snmaynard: Should probably be ~> 1.0

spec/session_tracker_spec.rb
```diff
@@ -3,13 +3,17 @@
 require 'spec_helper'
 require 'json'
 
+module Bugsnag
+  class SessionTracker
+    attr_accessor :track_sessions
```
snmaynard: What's this for?

spec/session_tracker_spec.rb
```diff
@@ -94,15 +102,21 @@
...
+      conf.auto_capture_sessions = true
       conf.release_stage = "test_stage"
       conf.delivery_method = :thread_queue
       conf.session_endpoint = "http://localhost:#{server.config[:Port]}"
     end
     WebMock.allow_net_connect!
-    Bugsnag.session_tracker.create_session
+    Bugsnag.session_tracker.start_session
+    while Bugsnag.session_tracker.session_counts.size == 0
+      sleep(0.05)
```
snmaynard: Why are these required?

spec/session_tracker_spec.rb
```diff
@@ -131,10 +145,10 @@
 
   it 'uses middleware to attach session to notification' do
     Bugsnag.configure do |conf|
-      conf.track_sessions = true
+      conf.auto_capture_sessions = true
       conf.release_stage = "test_stage"
     end
-    Bugsnag.session_tracker.create_session
+    Bugsnag.session_tracker.start_session
```
snmaynard: Can we just call `Bugsnag.start_session`

lib/bugsnag/session_tracker.rb
```diff
@@ -41,117 +36,93 @@ def create_session
...
-      end
-    end
+    alias :create_session :start_session
 
-    def deliver_sessions
-      return unless @config.track_sessions
+    def send_sessions
       sessions = []
-      @session_counts.each do |min, count|
+      counts = @session_counts.dup
```
snmaynard: You actually want

```ruby
counts = @session_counts
@session_counts = Concurrent::Hash.new(0)
counts.each ...
```

With the current implementation there is still a race condition.

lib/bugsnag/session_tracker.rb
```diff
@@ -41,117 +36,93 @@ def create_session
...
-      
-      if !@config.valid_api_key?
-        @config.debug("Not delivering sessions due to an invalid api_key")
+    private
+    def add_session(min)
+      @session_counts[min] += 1
+    end
+
+    def deliver(session_counts)
+      if session_counts.length == 0
```
snmaynard: be consistent about using `@session_counts` or `session_counts` in this file. I would use the member variable.

spec/session_tracker_spec.rb
Cawllec: It was being used to access a variable to stop the delivery threads being spun up, but is obsolete now.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ Changelog
 ### Enhancements
 
 * Adds support for tracking sessions and crash rate by setting the configuration option `configuration.track_sessions` to `true`.
-  Sessions can be manually created using `Bugsnag.start_session`, and manually delivered using `Bugsnag.send_sessions`.
+  Sessions can be manually created using `Bugsnag.create_session`, and manually delivered using `Bugsnag.send_sessions`.
```
snmaynard: Don't mention the `send_sessions` thing. Ideally that shouldn't even be public imo. It doesn't look like it's defined on Bugsnag anyway. The function should be `start_session` not `create_session`

Cawllec: Ah, that's the old changelog.  I'll fix the copy.

PR #415, opened 2018-01-10
Ensured sessions counts are preserved

Cawllec:
Ensures that event counts are preserved between multiple exceptions in a single session.

### DIFF
--- a/lib/bugsnag/middleware/session_data.rb
+++ b/lib/bugsnag/middleware/session_data.rb
@@ -12,6 +12,7 @@ module Bugsnag::Middleware
         else
           session[:events][:handled] += 1
         end
+        Bugsnag::SessionTracker.set_current_session session
         report.session = session
       end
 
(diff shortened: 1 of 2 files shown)


lib/bugsnag/middleware/session_data.rb
```diff
@@ -12,6 +12,7 @@ def call(report)
         else
           session[:events][:handled] += 1
         end
+        Bugsnag::SessionTracker.set_current_session session
```
snmaynard: I dont understand why this is needed

Cawllec: If multiple errors occur in a single session and are handled (i.e. not fatal) the session parameter should reflect that, showing the events incrementing over time.  Unless I'm fully misunderstanding how Ruby objects work, this should be necessary for that to happen.

snmaynard: You shouldnt need to set it again. You are just mutating the object

Cawllec: I didn't think that ruby objects worked like that, you're right, it's only PHP this is necessary.

PR #416, opened 2018-01-16
Fix shoryuken integration

smaximov:
This PR fixes shoryuken integration by using `Bugsnag.notify` instead of `Bugsnag.auto_notify` removed in a1b9e42dd. It also fixes a minor bug introduced in 7323b6a which causes Bugsnag to crash when `project_root` is `Pathname`, not a `String` (e.g., when `project_root` is set to `Rails.root` or `Hanami.root`).


Cawllec:
Ok, LGTM, thanks @smaximov

Issue #403, opened 2017-12-05
Rails: Middleware injection causes "can't modify frozen Array" error in test environment


sriedel:
I have a rails environment with a pretty standard bugsnag initializer:

     Bugsnag.configure do |config|
       config.api_key = "my_api_key_here"
       config.notify_release_stages = ['production', 'staging'] 
     end

As a testing environment I'm using rspec 3.6. 

When an error causes rspec to abort during parse-time, I get a plethora of "can't modify frozen Array" errors, obscuring the actual error causing the abort.

Adding a breakpoint to actionpack-5.0.6/lib/action_dispatch/middleware/stack.rb:96 (where the exception is ultimately thrown), I can see the following behavior:

- Rails loads its environment
- Middlewares are loaded
- the ruby interpreter tries to parse the test in question and fails, raising an exception
- Bugsnag tries to inject the Bugsnag::Rack middleware, causing the frozen Array error, since the middlewares array has been frozen after the Rails environment was done initializing
- The frozen array exception is printed to the console once spec file that was to be loaded

The obvious way to allevate the issue would be to disable the loading of the bugsnag middleware entirely when testing, or load the middlewares required during Rails' initialization. However I haven't found anything in the documentation of bugsnag how this could be accomplished.

I've tried adding config.middleware.disable( Bugsnag::Rack ) to the initializer, however this doesn't prevent the injection of the Bugsnag::Rack middleware.

Is there a way, other than to exclude bugsnag from the test group in bundler and wrap the initializer in a conditional checking the rails environment?

This is using bugsnag-ruby 5.4.1 (I know, not the most current version...)

Thanks!

Cawllec:
Hi @sriedel, at the moment we don't have a way of disabling specific middlewares during bugsnag initialisation.  I suspect your suggestion of excluding bugsnag from the test group would be the most effective way of stopping this behaviour, but I'll have a look at how you can stop the rack middleware loading and get back to you.

kattrali:
This will likely have the same solution as #391 (set an environment variable to prevent Bugsnag middleware from loading), though the intended release is a part of the 6.x series. I'll look into whether it would be simple to backport to 5.x as well.

sriedel:
Thanks for getting back on this, @kattrali. I've managed to upgrade to 6.1.1 in the meantime, and am using the Gemfile groups/intitializer workaround for now. So no need to backport on my account :)

cookrn:
[...]

PR #419, opened 2018-01-24
Added configuration option to disable autoconfiguration

### LINKED ISSUE #391: Difficulty configuring Bugsnag in Rails before it configures itself
Here's an example initializer in Rails:

``` ruby
# config/initializers/bugsnag.rb
puts "Bugsnag initializer reached"
Bugsnag.configure do |config|
  config.logger = Logger.new("/dev/null")
  config.api_key = ENV.fetch("BUGSNAG_API_KEY")
end
```

The Bugsnag ready message gets logged before this initializer runs:

```
$ rails c
** [Bugsnag] Bugsnag exception handler 5.5.0 ready
Bugsnag initializer reached
Loading development environment (Rails 5.1.4)
```

I want to change the Bugsnag logger and/or explicitly configure Bugsnag myself. I do not want the gem to try to automatically detect its own configuration. What is the correct way to do this in Rails and other Ruby environments when using Gemfiles/Rubygems to load this gem?

Cawllec:
This adds two functions, `load_integrations` & `load_integration(integration)` that allow Bugsnag integrations to be loaded during runtime.

By default, `load_integrations` will be run when the `bugsnag` module is required.  This behaviour can be disabled using the environment variable `BUGSNAG_DISABLE_AUTOCONFIGURE`.

Tentatively fixes #391, #403


snmaynard:
What manual testing has been done?

Cawllec:
I've tested it in rails to ensure the environment variable stops it from loading the railtie, which works correctly.  I guess the main issue would be that it still loads in `bugsnag\integrations\rake` regardless of environment variables as it's always done that separately to the other integrations.

snmaynard:
We should test it for all the integrations to ensure that manually loading the sidekiq integration (as an example) works and we dont need to make any changes to support the later loading.

Cawllec:
Now tested and confirmed to work with:
- Sidekiq
- Rails
- Resque

Does not work with:
- Rack

kattrali:
Rack or rake? Checking because rake was mentioned in https://github.com/bugsnag/bugsnag-ruby/pull/419#issuecomment-360204887

Cawllec:
Rack.  We loaded the Rack integration outside of the `load_integrations` loop previously, probably means removing it from there would make this a breaking change.

kattrali:
What broke when moving the import to the `load_integrations` loop?

Cawllec:
Nothing I suppose.  I haven't added it to the loop yet (as it was separate anyway), but it shouldn't change the expected behaviour if I do.  I'll do that and test it.

Cawllec:
Everything works as expected with rack, and there's no magic conf happening.

PR #420, opened 2018-01-26
Marks .bundle/ files out of project

Cawllec:
Marks any files in `.bundle/` as out of project along with `vendor/` files.


lib/bugsnag/stacktrace.rb
```diff
@@ -8,7 +8,7 @@ class Stacktrace
     JAVA_BACKTRACE_REGEX = /^(.*)\((.*)(?::([0-9]+))?\)$/
 
     # Path to vendored code. Used to mark file paths as out of project.
-    VENDOR_PATH = 'vendor/'
+    VENDOR_PATH = /^(vendor\/|.bundle\/)/
```
kattrali: `/^.bundle/` will match "abundle" or "5bundle" as full stop is a wildcard expression for any character. Should be escaped with a backslash.

Issue #423, opened 2018-02-01
Incorrect interpretation of proxy port numbers


cluesque:
### Expected behavior
When `ENV['http_proxy']` is of the form `http://192.168.1.2:8888` the port number should be respected.

### Observed behavior
Bug reports don't escape because an attempt is made to connect to `192.168.1.2:8888:80`, as reported by the sad proxy.

### Steps to reproduce
Run a rails app in an environment where `ENV['http_proxy']` is of the form `http://192.168.1.2:8888`.
Force a reportable exception.

### Version
6.6.3

### Additional information
We discovered this when upgrading from version 5.4.1

PR #424, opened 2018-02-01
If ENV['http_proxy'] has port, use it

cluesque:
... also allow for user and password, if configured

fixes bugsnag/bugsnag-ruby#423


Cawllec:
LGTM

Issue #422, opened 2018-02-01
bugsnag_meta_data regression


abraham-chan:
### Expected behavior
Copy pasting the example from [the docs](https://docs.bugsnag.com/platforms/ruby/rails/customizing-error-reports/#custom-error-diagnostic-data) into a console on v5.4.1 results in the tab appearing correctly on Bugsnag.

### Observed behavior
After updating to v6.6.3 the same code does not create the tab.

### Steps to reproduce
```
class MyCustomError < StandardError
  attr_reader :metadata
  def initialize(message, metadata)
    super(message)
    @metadata = metadata
  end

  def bugsnag_meta_data
     {tabname: metadata}
  end
end

Bugsnag.notify(MyCustomError.new("Error message", value1: '1', value2: {nested: 3}))
```

### Version
6.6.3

Cawllec:
Thanks for the info @abraham-chan, this appears to be a case of the docs lagging behind a little bit.  In version 6 of the notifier there's a metadata class `Bugsnag::MetaData` that should be included into the custom error in order for the middleware to extract metadata.
The example above should actually be:
```
class MyCustomError < StandardError
  include Bugsnag::MetaData
  attr_reader :metadata
  def initialize(message, metadata)
    super(message)
    @metadata = metadata
  end

  def bugsnag_meta_data
     {tabname: metadata}
  end
end

Bugsnag.notify(MyCustomError.new("Error message", value1: '1', value2: {nested: 3}))
```
I'll make sure I get the docs update right away

kattrali:
Hmm, this seems like a mistake though. There's no need to require a module to make this work, as we can just check if the exception responds to `bugsnag_meta_data`.

Cawllec:
In that case I'll create a fix for this to take it back to the v5 behaviour.

kattrali:
Fixed in [v6.6.4](https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.6.4)

PR #425, opened 2018-02-05
Calls bugsnag_* methods on custom exceptions if available

Cawllec:
Based on feedback in #422 and discussion with @kattrali, this PR makes it unnecessary for a custom exception to have included `Bugsnag::MetaData` by checking the functions exist individually.

### DIFF
--- a/lib/bugsnag/middleware/exception_meta_data.rb
+++ b/lib/bugsnag/middleware/exception_meta_data.rb
@@ -7,23 +7,21 @@ module Bugsnag::Middleware
     def call(report)
       # Apply the user's information attached to the exceptions
       report.raw_exceptions.each do |exception|
-        if exception.class.include?(Bugsnag::MetaData)
-          if exception.bugsnag_user_id.is_a?(String)
-            report.user = {id: exception.bugsnag_user_id}
-          end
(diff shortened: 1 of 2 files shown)


lib/bugsnag/middleware/exception_meta_data.rb
```diff
@@ -7,23 +7,21 @@ def initialize(bugsnag)
...
-          end
+        if exception.respond_to?(:bugsnag_grouping_hash) && exception.bugsnag_grouping_hash.is_a?(String)
+          report.grouping_hash = exception.bugsnag_grouping_hash
+        end
 
-          if exception.respond_to?(:bugsnag_meta_data) && exception.bugsnag_meta_data
-            exception.bugsnag_meta_data.each do |key, value|
-              report.add_tab key, value
-            end
+        if exception.respond_to?(:bugsnag_meta_data) && exception.bugsnag_meta_data
```
snmaynard: It would seem wise to add a type check instead of just a presence check here (like we have in the other if statements)

Cawllec:
Updated this with tests that verify the `bugsnag_*` methods are still checked on exceptions that don't mixin `Bugsnag::MetaData`.

Cawllec:
Tests appear to be failing on this due to travis issues with Jruby.  It installs and runs through tests fine locally.

kattrali:
Closed in favor of #426

PR #431, opened 2018-02-21
WIP: Up limits on payload lengths, more focussed trimming

Cawllec:
Changes how payload trimming is implemented to favour keeping longer stacktraces in place rather than limiting them to 40 items.

Namely:
- Ups max payload length to 512000 bytes (~0.5 Mb)
- Ups max metadata array length to 80
- Ups max other array length (namely stacktraces) to 160
- Changes order of payload trimming to:
  1. Trimming strings to max length (3072)
  2. Trimming metadata arrays to max length (80)
  3. Trimming all other arrays (including stacktraces) to max length (160)
  4. Removing metadata from the payload
  5. Remove `:code` from stacktrace items, oldest first, until within limits
  6. Remove stacktrace items, oldest first, until within limits

WIP:
1. I need to write performance tests to compare with the older version to see if extra performance time is within acceptable limits
2. Review of how iteration is performed over payload to ensure performing optimally
3. Review of tests to ensure all cases are covered


lib/bugsnag/stacktrace.rb
```diff
@@ -1,3 +1,5 @@
+require 'pp'
```
kattrali: Is this from debugging?

spec/helper_spec.rb
```diff
@@ -171,6 +172,111 @@ def to_s
           expect(trimmed_value[:events].first[:metaData]).to be_nil
         end
       end
+
+      context "and new trimming priorities are in place" do
```
kattrali: This shouldn't be a new context because no conditions are being changed.

lib/bugsnag/helpers.rb
```diff
@@ -60,6 +83,45 @@ def self.deep_merge!(l_hash, r_hash)
 
     TRUNCATION_INFO = '[TRUNCATED]'
 
+    # Trim stacktrace code out if they're too large, oldest functions first
+    def self.trim_stacktrace_code(payload, threshold)
+      return payload unless payload.is_a?(Hash) and payload[:events].respond_to?(:map)
+      payload[:events].map do |event|
+        event[:exceptions].map do |exception|
+          initial_size = get_payload_length(exception[:stacktrace])
```
kattrali: just do `exception[:stacktrace].each` and pop the code from each entry. No need for all of these checks and breaks or using `downto`.

lib/bugsnag/helpers.rb
```diff
@@ -60,6 +83,45 @@ def self.deep_merge!(l_hash, r_hash)
...
+          (exception[:stacktrace].length - 1).downto(0).each do |i|
+            break unless (initial_size - get_payload_length(exception[:stacktrace])) < threshold
+            exception[:stacktrace][i].delete(:code) if exception[:stacktrace][i].include?(:code)
```
kattrali: No need to check for inclusion before deletion.

lib/bugsnag/helpers.rb
```diff
@@ -60,6 +83,45 @@ def self.deep_merge!(l_hash, r_hash)
...
+        end
+      end
+      payload
+    end
+
+    # Trim stacktrace entries out oldest functions first
+    def self.trim_stacktrace_functions(payload, threshold)
+      return payload unless payload.is_a?(Hash) and payload[:events].respond_to?(:map)
+      payload[:events].map do |event|
+        event[:exceptions].map do |exception|
```
kattrali: Same changes as above ☝️

lib/bugsnag/stacktrace.rb
Cawllec: Yes, I'll remove that.

spec/helper_spec.rb
Cawllec: It's mostly a readability context for me at the moment before the tests are concreted.

lib/bugsnag/helpers.rb
Cawllec: :+1:

snmaynard:
Can you ping me on this when it's ready for a second review.

lib/bugsnag/helpers.rb
```diff
@@ -60,21 +70,74 @@ def self.deep_merge!(l_hash, r_hash)
...
     # Check if a value is a raw type which should not be trimmed, truncated
     # or converted to a string
     def self.is_json_raw_type?(value)
       RAW_DATA_TYPES.detect {|klass| value.is_a?(klass)} != nil
     end
 
+    ##
     # Shorten array until it fits within the payload size limit when serialized
-    def self.truncate_array(array)
+    def self.truncate_array(array, limit = MAX_ARRAY_LENGTH)
```
kattrali: If `limit` is just a constant value everywhere and never modified, just use it directly instead of passing it from one method to the next.

PR #433, opened 2018-02-23
Add ruby & framework versions to app parameter

Cawllec:
Adds `RUBY_VERSION` and select framework version numbers to an events `app` parameter.

Covered frameworks are:
- rack
- rails
- que
- rake
- sinatra
- sidekiq

App parameter example:
```
app: {
          version: null,
          releaseStage: 'development',
          type: 'rails',
          RUBY_VERSION: '2.4.1',
          rackVersion: '1.3',
          railsVersion: '5.1.4' 
}
```

Manual testing performed on:
- [x] Vanilla
- [x] Que
- [x] Rack
- [x] Rails
- [x] Sinatra
- [x] Sidekiq
- [x] Rake


.rubocop_todo.yml
```diff
@@ -288,6 +288,8 @@ Metrics/BlockNesting:
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
   Max: 149
+  Exclude:
+    - 'lib/bugsnag/report.rb'
```
kattrali: Don't add new exceptions to the linter.

snmaynard:
Can we make it rubyVersion please

.rubocop_todo.yml
```diff
@@ -287,7 +287,7 @@ Metrics/BlockNesting:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 149
+  Max: 150
```
snmaynard: Why is this in this PR?

spec/integrations/que_spec.rb
```diff
@@ -39,6 +40,9 @@ class ::Que
         :framework => 'Que'
       }
     })
+    expect(report).to receive(:add_tab).with(:app, {
+      :queVersion => ::Que::Version
```
snmaynard: this should be the literal 'test'

spec/integrations/rails3_request_spec.rb
```diff
@@ -3,6 +3,7 @@
 describe Bugsnag::Middleware::Rails3Request do
   before(:each) do
     Bugsnag.configuration.middleware.use(described_class)
+    stub_const('::Rails::VERSION::STRING', 'test_version')
```
snmaynard: Why do we use this for Rails and define a const for Que?

.rubocop_todo.yml
Cawllec: As the class required an extra line (the `RUBY_VERSION` line in the payload) and we didn't feel arbitrary limits disallowing new lines to be added (or requiring other lines be removed) without proper thought into refactoring is a good idea.

spec/integrations/rails3_request_spec.rb
Cawllec: This should be used where possible.  I've updated the Que version, and although that still requires a sample module for mocking purposes, it simplifies the tests.

snmaynard:
Closing this as it will be part of a broader project to roll this out across all notifiers

Issue #417, opened 2018-01-16
before_notify_callbacks don't work with Sidekiq


tsujigiri:
### Expected behavior
When an exception is raised in a Sidekiq worker, it should also run the `before_notify_callbacks`.

### Observed behavior
It doesn't. 🤷

### Steps to reproduce

I set up a new Rails app and added sidekiq and bugsnag: https://github.com/ad2games/sidekiq_bugsnag_test

* Before running, the `BUGSNAG_API_KEY` environment variable needs to be set.
* Sidekiq needs to be started: `bundle exec sidekiq`
* When `Bugsnag.notify(StandardError.new('foo'))` is being run from the console directly, we end up in the callback.
* When `FooWorker` is run via `FooWorker.perform_async` from the Rails console, the callback in the initializer is not called.

### Version

See [Gemfile.lock](https://github.com/ad2games/sidekiq_bugsnag_test/blob/master/Gemfile.lock)

Cawllec:
Hi @tsujigiri, the reason is because we store the `request_data` (including the callbacks) in `Thread.current`, meaning when Sidekiq spawns a thread to run `FooWorker` and the `request_data` is attached to the `report` it won't be the same array.

You can see this from adding checks before the callback is registered and in the worker.
The results I got were:
`"#<Thread:0x007fbfe007ef70>:Registering callback"`
and:
`"#<Thread:0x007fbfe0ab5a70>:PERFORMING TASK"`.
I'm not entirely sure how you'd fix this aside from running `before_notify_callbacks <<` in the worker itself.  We'll have to look at how the client can deal with storing this data across threads.

tsujigiri:
Why is it necessary to store the callbacks in the thread? They shouldn't ever change other than as part of the initial configuration, right?

Cawllec:
They allow you to add callbacks scoped to a particular request, which won't be retained when the thread closes, stopping memory leaks and retention of things like user information in memory.

I'll make sure the need to initialise the callback in the task rather than the server is documented properly.

tsujigiri:
This seems like a bit of an edge-case to me and it makes everyone having to make sure every thread has the callbacks set, that are supposed to always be called. If you think thread-local callbacks are really necessary, maybe having both options, global callbacks as well as thread-local ones, would be a good compromise?

Cawllec:
[...]

PR #434, opened 2018-02-26
Updates to use sidekiq v3 error handling

Cawllec:
Ensures that an error handler is correctly assigned in newer versions of sidekiq.

While it has very basic sidekiq integration tests already, this is a prime candidate for the new integrations test runner, which I will be working with this week.

Manually tested in:

- [x] Sidekiq 2
- [x] Sidekiq 3


snmaynard:
I would like to see integration tests for this, rather than just unit. Can we add some maze tests?

javierjulio:
Will this address #417 too or is that being done in another PR?

Cawllec:
That will be in a separate PR, as decisions about that change still have to be made.

snmaynard:
I have reworked the approach here to give us the sidekiq queue information in all versions. I ran the maze tests and they all passed.

PR #436, opened 2018-03-02
Add warden.user rack data to default filters list

Cawllec:
Ensures rack data containing warden user id's and partial encrypted password hashes will be correctly filtered by default.

Warden data will still be available as metadata added via the Warden user middleware.

### DIFF
--- a/spec/report_spec.rb
+++ b/spec/report_spec.rb
@@ -574,7 +574,27 @@ describe Bugsnag::Report do
   it "filters params from all payload hashes if they are set in default meta_data_filters" do
 
     Bugsnag.notify(BugsnagTestException.new("It crashed")) do |report|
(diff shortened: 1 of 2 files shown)


snmaynard:
We would whitelist the fields in the session tab and then if warden changes we would be alerted.

spec/report_spec.rb
```diff
@@ -574,7 +574,27 @@ def gloops
...
+          },
+          :cookie => "1234567890",
+          :authorization => "token",
+          :user_authorization => "token",
+          :secret_key => "key",
+          :user_secret => "key"
+        }
+      })
+      report.meta_data.merge!({
+        :wardenUser => {
```
snmaynard: Can we put this in a representative place in the payload?

PR #437, opened 2018-03-07
Expanded support for proxy config from environment variables

Cawllec:
Expands setting of `config.proxy` options to both `ENV['http_proxy']` and `ENV['https_proxy']` automatically.

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -77,7 +77,7 @@ module Bugsnag
 
       # Read NET::HTTP proxy environment variables
       if proxy_uri = ENV["https_proxy"] || ENV['http_proxy']
-        set_proxy(proxy_uri)
+        parse_proxy(proxy_uri)
       end
 
       # Set up logging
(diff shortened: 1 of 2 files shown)


Cawllec:
Docs PR incoming for this as well.

Cawllec:
@kattrali can I get a review on this for merging?

lib/bugsnag/configuration.rb
```diff
@@ -184,6 +180,16 @@ def debug(message)
       logger.debug(PROG_NAME) { message }
     end
 
+    ##
+    # Parses and sets proxy from a uri
+    def parse_proxy(uri)
+      proxy = URI.parse(uri)
```
fractalwrench: Would there be any problems if `uri` was nil at this point?

Cawllec: Absolutely, however this would only occur if `parse_proxy` was called with `nil`, in which case `URI` would spit out a legible error i.e. `**URI::InvalidURIError: bad URI(is not URI?):**`

Cawllec: I could catch and log that error, but I think failing loudly is more valuable in most cases.

PR #439, opened 2018-03-19
Handle 'nil' being notified

Cawllec:
- Changes 'nil' to a string: "'nil' was notified as an exception"
- Fixes Bugsnag throwing exception during notification process


spec/report_spec.rb
```diff
@@ -1045,6 +1045,15 @@ def gloops
     expect(Bugsnag).not_to have_sent_notification
   end
 
+  it 'uses an appropriate message if nil is notified' do
+    Bugsnag.notify(nil)
+    expect(Bugsnag).to have_sent_notification{ |payload, headers|
+      event = payload["events"][0]
+      exception = event["exceptions"][0]
+      expect(exception["errorClass"]).to eq("RuntimeError")
+      expect(exception["message"]).to eq(Bugsnag::Report::NIL_EXCEPTION_DESCRIPTION)
```
kattrali: What does the stacktrace look like in this case? It would be good to verify that it is also sensible in a test.

Cawllec:
This is waiting for the refactoring change to be merged first #440

snmaynard:
Why does this depend on the refactoring?

Why was this not done in the Bugsnag Client? We already have a bunch of validation of input in there.

spec/report_spec.rb
Cawllec: I've updated the test to look at the stacktrace as well now.

snmaynard:
I still have a question about this being done in report rather than in the client.

Cawllec:
It makes sense to me that the `notify` function handles whether a notification _should_ be sent, whereas the report class handles the content of the notification.  As this falls into the latter category, I put it in the Report.

snmaynard:
The notify method is the public interface, and so it validates user input - it should be there.

Cawllec:
I've moved the nil exception changes into the Client, and refactored the configuration checks into a private `deliver_notification?` method.  This negates the need for the refactor PR, which I'll close.  I'll put together a plan for a proper refactoring that will hit more than a couple of points which can be discussed at a later date.

PR #441, opened 2018-03-21
WIP: Plain Maze tests

Cawllec:
Adds basic level Maze tests for vanilla Ruby
- Covers the initial list of basic tests
- Creates foundation for more advanced tests


Cawllec:
Note: The reason this is failing is that Bundler attempts to build a gemfile.lock by searching through all of the listed Gems, even if they're in a group that's not being installed.  This can be fixed by adding a `--deployment` flag, although how this will effect the installation I don't know.

Cawllec:
I've run the full test suite locally and fixed a bunch of the issues that were cropping up. If I can get confirmation on that, I'd be happy to add CI back in through a separate PR.

snmaynard:
I dont think adding maze to CI is going to be viable as I think travis will time out the suite. We could add a shortened version to CI and then run the full suite before doing a release or something?

Cawllec:
When running each feature folder on a separate job it ran in under the time.  Otherwise we could run priority tests in plain_features and run the rest manually when necessary

snmaynard:
I'd rather the public tests ran in sub 10 minutes I think. How long did it take?

Cawllec:
A little longer than that, around 45 minutes (for only the plain tests).  This could potentially be reduced, but would need a reworking of the docker approach and would be a reasonably significant change.

I'm happy to run them manually for the time being, as we can select which tests need to be run to manually test a given feature.  The full suite is currently clocking in ~2 hours.

PR #443, opened 2018-03-26
feature(logging): Ensure logged messages are formatted correctly

Cawllec:
- Ensures logged messages have the correct bugsnag formatting
- Ensures custom loggers receive the formatted message rather than the raw message
- Adds a new public method - 'format_message'

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -185,7 +185,7 @@ module Bugsnag
     # Formats a message being logged by Bugsnag
     def format_message(message)
       datetime = Time.new.to_s
-      "** [Bugsnag] #{datetime}: #{message.to_s}\n"
+      "** [Bugsnag] #{datetime}: #{message}\n"
     end
 
     private
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -88,9 +88,6 @@ module Bugsnag
       # Set up logging
       self.logger = Logger.new(STDOUT)
       self.logger.level = Logger::INFO
-      self.logger.formatter = proc do |severity, datetime, progname, msg|
-        "** [Bugsnag] #{datetime}: #{msg}\n"
-      end
 
       # Configure the bugsnag middleware stack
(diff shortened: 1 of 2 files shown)


lib/bugsnag/configuration.rb
```diff
@@ -88,9 +88,6 @@ def initialize
       # Set up logging
       self.logger = Logger.new(STDOUT)
       self.logger.level = Logger::INFO
-      self.logger.formatter = proc do |severity, datetime, progname, msg|
-        "** [Bugsnag] #{datetime}: #{msg}\n"
```
kattrali: Keep the formatter as before, only removing the `[Bugsnag]` component

lib/bugsnag/configuration.rb
```diff
@@ -169,19 +166,26 @@ def clear_request_data
     ##
     # Logs an info level message
     def info(message)
-      logger.info(message)
+      logger.info(format_message(message))
```
kattrali: Use the [block form of Logger logging methods](https://ruby-doc.org/stdlib-1.9.2/libdoc/logger/rdoc/Logger.html#method-i-info) to specify the program name which sent the message. This differentiates Bugsnag log messages from others sent using the same logger.

```diff
- logger.info(format_message(message))
+ logger.info('Bugsnag') { message }
```

lib/bugsnag/configuration.rb
```diff
@@ -169,19 +166,26 @@ def clear_request_data
...
     ##
     # Logs a debug level message
     def debug(message)
-      logger.debug(message)
+      logger.debug(format_message(message))
+    end
+
+    ##
+    # Formats a message being logged by Bugsnag
+    def format_message(message)
```
kattrali: While this is unneeded in a simpler implementation, it should be noted that this method should have been private as it has no role in the public interface.

Cawllec:
I've adjusted how this works based on your feedback - the formatters back in place and the `PROG_NAME` const will be delivered as the program name in all logger cases.

kattrali:
Looks good, I made one adjustment to make the `PROG_NAME` constant private and title-case it as before.

PR #444, opened 2018-03-31
fix(rails): Log missing key warning after Rails initialization completes

### LINKED ISSUE #391: Difficulty configuring Bugsnag in Rails before it configures itself

kattrali:
When not initializing using the `BUGSNAG_API_KEY` environment variable,
the first time the Railtie integration is loaded, no API key is 
available. At the bottom of the configure method, a warning is emitted
stating that no API key has been set, though this may not be true after
`config/initializers/bugsnag.rb` runs. This is the behavior being observed
in #391, causing concern that bugsnag is not working correctly when in
fact the log message is premature.

This change allows for `Bugsnag.configure` to be called without validating
the key, allowing the Railtie integration to add an additional check
once Rails initialization is complete.

Fixes #391

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -37,14 +37,10 @@ module Bugsnag
     # Configure the Bugsnag notifier application-wide settings.
     #
     # Yields a configuration object to use to set application settings.
-    def configure
+    def configure(validate_api_key=true)
       yield(configuration) if block_given?
 
-      @key_warning = false unless defined?(@key_warning)
-      if !configuration.valid_api_key? && !@key_warning
-        configuration.warn("No valid API key has been set, notifications will not be sent")
-        @key_warning = true
-      end
+      check_key_valid if validate_api_key
     end
 
     ##
@@ -177,6 +173,15 @@ module Bugsnag
         require "bugsnag/integrations/#{integration}"
       else
         configuration.debug("Integration #{integration} is not currently supported")
+      end
+    end
+
+    # Check if the API key is valid and warn (once) if it is not
+    def check_key_valid
+      @key_warning = false unless defined?(@key_warning)
+      if !configuration.valid_api_key? && !@key_warning
+        configuration.warn("No valid API key has been set, notifications will not be sent")
+        @key_warning = true
       end
     end
   end
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -36,7 +36,7 @@ module Bugsnag
 
     config.before_initialize do
       # Configure bugsnag rails defaults
-      Bugsnag.configure do |config|
+      Bugsnag.configure(false) do |config|
         config.logger = ::Rails.logger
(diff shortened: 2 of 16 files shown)


lib/bugsnag/integrations/railtie.rb
```diff
@@ -69,6 +69,7 @@ class Railtie < Rails::Railtie
           end
         end
       end
+      Bugsnag.configure(true) # Validate API key is set
```
snmaynard: You dont even need this because the previous configure will also validate?

kattrali: There isn't another configure call if the generator was never run.

snmaynard: The one above I mean?

kattrali: Oh you mean the immediately preceding line? Yes, that would be more logical.

lib/bugsnag/integrations/railtie.rb
```diff
@@ -36,7 +36,7 @@ class Railtie < Rails::Railtie
 
     config.before_initialize do
       # Configure bugsnag rails defaults
-      Bugsnag.configure do |config|
+      Bugsnag.configure(false) do |config|
```
snmaynard: get rid of the (false)?

spec/integrations/logger_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    context 'skips setting an API key' do
+      it 'logs a warning' do
+        skip "Incompatible with Ruby <2.0 and JRuby" if incompatible
+        output = run_app('rails-no-config')
+        expect(output).to include(key_warning)
+      end
+    end
+  end
```
martin308: a test in here where the env variable is provided but is not a valid api key and `configure` is not called might be a good addition

lib/bugsnag.rb
```diff
@@ -37,14 +37,10 @@ class << self
     # Configure the Bugsnag notifier application-wide settings.
     #
     # Yields a configuration object to use to set application settings.
-    def configure
+    def configure(validate_api_key=true)
```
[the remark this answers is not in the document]
kattrali: Supporting Ruby 1.9.3 makes this difficult as 2.0.0 introduced named parameters. It could potentially be something for the next major release.

PR #446, opened 2018-04-02
Fix max_attempts notification for delayed_job

calavera:
Custom jobs can override the default max attempts. Bugsnag doesn't take that into account.

See: https://github.com/collectiveidea/delayed_job/blame/ccbde3dbcd90a409fe3dceae906fa6292ef09692/lib/delayed/worker.rb#L288


Cawllec:
Hi @calavera, thanks for the PR, looks good to me - and I'm going to use this as an opportunity to revisit the delayed-job unit tests in Bugsnag.

PR #449, opened 2018-04-08
Allow inProject stacktrace filters to be configured

phil-monroe:
The `vendor` and `.bundle` filters work great for gem code, but we have some helper classes in our code base to DRY up common notification patterns that we would like to be ignored from the Bugsnag stacktraces and groupings.

This PR allows additional stacktrace filters to be added like so:

``` ruby
Bugsnag.configure do |config|
  config.stacktrace_filters << %r{^app/services/exception_notifier.rb}
end
```

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -22,6 +22,7 @@ module Bugsnag
     attr_accessor :app_version
     attr_accessor :app_type
     attr_accessor :meta_data_filters
+    attr_accessor :stacktrace_filters
     attr_accessor :endpoint
     attr_accessor :logger
     attr_accessor :middleware
@@ -51,6 +52,12 @@ module Bugsnag
       "rack.request.form_vars"
     ].freeze
 
+    DEFAULT_STACKTRACE_FILTERS = [
+      # Path to vendored code. Used to mark file paths as out of project.
+      /^vendor\//,
+      /^\.bundle\//
+    ]
+
     alias :track_sessions :auto_capture_sessions
     alias :track_sessions= :auto_capture_sessions=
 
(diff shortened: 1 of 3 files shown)


Cawllec:
Hi @phil-monroe, thanks for the contribution.  We typically try to minimise the amount of changes to configuration options to maintain parity across our notifiers, but we'll review this change and see if it's something we think could be beneficial.

In the meantime, you can edit your stacktraces in middleware or callbacks:
```
Bugsnag.before_notify_callbacks << Proc.new do |report|
  report.exceptions[0][:stacktrace].delete_if do |trace|
    /^app\/services\/exception_notifier.rb$/.match(trace[:file])
  end
end
```

Cawllec:
Hi @phil-monroe, I may have misinformed you in my above answer.  What you seem to be wanting to do is to set `inProject` on the stacktrace using the same logic.

```
Bugsnag.before_notify_callbacks << Proc.new do |report|
  report.exceptions[0][:stacktrace].each do |trace|
	trace[:inProject] = false if /^app\/services\/exception_notifier.rb$/.match(trace[:file])
  end
end
```

phil-monroe:
Ah, neat! Didn't realize we had access to `inProject` in `before_notify_callbacks `. Thanks!

phil-monroe:
We ended up going with a middleware approach over `before_notify_callbacks` since `before_notify_callbacks` does not work with threaded applications (sidekiq - #417)

Here is our solution for reference:

``` ruby
class BugsnagIgnoreProjectLines
  def initialize(bugsnag)
    @bugsnag = bugsnag
  end

  def call(report)
      exception[:stacktrace].each do |trace|
        trace[:inProject] = false if trace[:file] =~ %r{^app/services/maven/notifier.rb$}
      end
    end
  end
end


Bugsnag.configure do |config|
  config.middleware.use(BugsnagIgnoreProjectLines)
end
```

PR #452, opened 2018-04-18
tests(CI): Add travis job pattern for plain-ruby maze tests

Cawllec:
Verified running on forked project.

Note: Squash before merging


.travis.yml
```diff
@@ -56,6 +56,19 @@ jobs:
...
+        packages:
+        - docker-ce
+    sudo: required
+    sevices:
+    - docker
+    env:
+    - BUNDLE_VERSION=1.12.0
+    - GEMSETS="test maze"
+    rmv: 2.4.1
+    script: bundle exec bugsnag-maze-runner features/plain_features
```
fractalwrench: Do we only want to run this particular feature on CI, or are there others?

features/support/env.rb
```diff
@@ -1,5 +1,7 @@
 require 'open3'
 
 def current_ip
-  `ifconfig | grep -Eo 'inet (addr:)?([0-9]*\\\.){3}[0-9]*' | grep -Eo '([0-9]*\\\.){3}[0-9]*' | grep -v '127.0.0.1'`.strip
+  ip_addr = `ifconfig | grep -Eo 'inet (addr:)?([0-9]*\\\.){3}[0-9]*' | grep -v '127.0.0.1'`
+  ip_list = /((?:[0-9]*\.){3}[0-9]*)/.match(ip_addr)
+  ip_list.captures.first
 end
```
fractalwrench: \n

features/support/env.rb
```diff
@@ -1,5 +1,7 @@
 require 'open3'
 
 def current_ip
```
fractalwrench: A comment might be useful here - does this just parse the output of `ifconfig` to obtain the current IP address, or are there additional steps?

.travis.yml
Cawllec: I'm splitting it into jobs, as the 50minute limit is a concern if the plain and rails tests were to run serially

Cawllec: I'll update `.travis.yml` in the rails PR to add that job

features/support/env.rb
Cawllec: I've added a comment - hopefully the rest should be self-explanatory

PR #453, opened 2018-04-18
feature(Delayed job): Use overridden :max_attempts

Cawllec:
Based on [this PR](https://github.com/bugsnag/bugsnag-ruby/pull/446).

- Adds `:max_attempts` override for metaData
- Expands unit test to cover correct payload construction (including `:max_attempts`)


Cawllec:
Flaky test - I've moved it into maze for the future, but until that's merged this will probably keep reoccurring.

lib/bugsnag/integrations/delayed_job.rb
```diff
@@ -30,7 +30,8 @@ def error(job, error)
               overrides[:job][:queue] = queue
             end
             if job.respond_to?(:attempts)
-              overrides[:job][:attempts] = "#{job.attempts + 1} / #{Delayed::Worker.max_attempts}"
+              max_attempts = job.max_attempts || Delayed::Worker.max_attempts
```
kattrali: Shouldn't this `respond_to?` check whether the job responds to `max_attempts` as well as `attempts`?

PR #460, opened 2018-05-05
[PLAT-1079] apply meta_data_filters to referer 

tremlab:
apply metadata filters to referer and headers Referer to redact any values that should be filtered from query params (identical to url cleaning).  Added test for redaction in url, referer and headers Referer.


Cawllec:
Looks good, I've switched the target branch to `next`, and there are a couple of fixes for the Rack middleware tests that should be merged from there to this branch.  If you could do that it'll be easier to tell if the requests pass.

kattrali:
Moved to #468, did a bit of commit pruning.

PR #466, opened 2018-05-14
Cawllec/update rake version

Cawllec:
## Goal

Allow new versions of Rake to be used with testing and later Ruby versions, while retaining

### Changed

- Added conditional to Rake to ensure compatibility across Ruby versions 

## Tests

Locally tested the correct Rake versions are loaded on 1.9.3 vs newer Ruby versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


Cawllec:
We still need it for 1.9.3 unfortunately

PR #467, opened 2018-05-17
v6.8.0

Cawllec:
Merge in for v6.8.0 release


Cawllec:
Ah, yeah, this was meant to be a fix release for the rake version, but we don't need that released.  Sorry about that.

PR #472, opened 2018-06-11
fix: Remove local rails config

Cawllec:
## Goal

The example config used for local testing had been checked in at some point, this rectifies that.


example/rails-51/config/initializers/bugsnag.rb
```diff
@@ -1,5 +1,4 @@
 Bugsnag.configure do |config|
-  config.api_key = "12312312312312312312312312312312"
-  config.auto_capture_sessions = false
-  config.endpoint = "http://localhost:62000"
+  config.api_key = "YOUR_API_KEY_HERE"
+  config.auto_capture_sessions = true
```
martin308: this looks to have flipped to true, should that be the case?

PR #473, opened 2018-06-16
Add gem version and SemVer compatibility badges to README

greysteil:
## Goal

Make it easier for new users to add `bugsnag` to their Gemfile.

## Design

Badges that show the version and how SemVer compatible this gem is (very) give users all the information they need to pick the right gem requirement.

This PR adds a version number and SemVer compatibility badge. Data comes from Dependabot (disclosure: I built it).

Here's what the badges look like:

Gem Version [SemVer](https://dependabot.com/compatibility-score.html?dependency-name=bugsnag&package-manager=bundler&version-scheme=semver)

## Review

  - [ ] Final review


greysteil:
Yep - the idea here is to test SemVer by checking whether updates to bugsnag cause test failures for your users. The "score" is just the percentage of CI runs that pass when updating between SemVer compatible versions, but it still gives a pretty clear indicator that bugsnag updates won't require code changes for your users if/when the upgrade it.

There are some difficulties with flaky tests that I haven't 100% ironed out yet (I should be able to filter out flaky test suites based on all the other updates Dependabot is doing for that repo) but in practice it seems flaky tests only affect the score very slightly. Of the 163 updates Dependabot has done of bugsnag only 3 have ever "caused" test failures.

snmaynard:
Yeah - I just don't think that a 98% score is particularly clear to people who don't know what it is and how it works. I think that we'll likely get confused users if we merge this. Maybe you could give a grading system instead, but as it is I dont think it's clearer than us saying we use semver versioning.

greysteil:
👍, although sorry to hear that. I've renamed the badge to be "semver stability" to make it clearer what it does, and I'm going to try to filter out the flaky specs later today / tomorrow.

Keep up the great work with Bugsnag! ❤️

snmaynard:
You could do something like 95%+ is given an excellent rating?

snmaynard:
Yeah fair - do you have examples of other projects that use it?

greysteil:
I only created it a few days, but [jest](https://github.com/facebook/jest#jest) have added it, [sentry](https://github.com/getsentry/raven-ruby/#raven-ruby-the-ruby-client-for-sentry) have, [capybara](https://github.com/teamcapybara/capybara#capybara), [oj](https://github.com/ohler55/oj#-gem), and [rubocop](https://github.com/rubocop-hq/rubocop) (in prep for going 1.0.0 in the next few weeks). Think there's a couple of others, too.

snmaynard:
I'll reopen it and see what @kattrali thinks when she is back from pto next week - thanks for all the context!

greysteil:
@kattrali - any thoughts? The folks at Webpack asked for a version that only displayed updates to the latest version, so there's this option now, too:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=bugsnag&package-manager=bundler&version-scheme=semver&new-version=latest)

kattrali:
This is pretty cool! The context discussion helps too, so thanks @greysteil, @snmaynard.

PR #475, opened 2018-07-05
fix: Add mac specific host for docker

Cawllec:
Stops ip issues when communicating to the WeBrick server from inside the maze docker container, when running locally.


snmaynard:
In what way is this not working currently?

Cawllec:
When running locally the IP obtained by the original function hasn't been accessible from within the container.  Even if this is only a quirk of a docker version, or my current machine, this allows us to ensure a mac-specific hostname is set, and will allow us to expand it to work on windows.

PR #477, opened 2018-07-20
Do not report SignalExceptions in at_exit handler

mrnugget:
## Goal

The previous behaviour resulted in applications reporting errors even though they were gracefully and without any problems shut down. Example: the puma webserver relies on `SIGTERM` to gracefully shut down its workers and Bugsnag reported errors when it was normally shut down.

With the `at_exit` handler introduced in #397, Bugsnag would catch `SignalExceptions` on exit and report them as error, even though they are none. 

This PR changes the existing behaviour to add `SignalException` as a special case in which the exception should not be passed to `Bugsnag.notify`.

### Added

- `Bugsnag.at_exit_handler`

### Changed

- `Bugsnag.register_at_exit`

## Tests

Due to the Ruby exiting when a `SignalException` is raised I had to refactor the code to make the actual handling of the exit-exceptions testable without the test suite quitting.

### Linked issues

For more information on puma's behaviour, see here: https://github.com/puma/puma/issues/1438
Honeybadger also ignores `SignalException`s: https://github.com/honeybadger-io/honeybadger-ruby/issues/267

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -118,15 +118,17 @@ module Bugsnag
     def register_at_exit
       return if at_exit_handler_installed?
       @exit_handler_added = true
-      at_exit do
-        if $!
-          Bugsnag.notify($!, true) do |report|
-            report.severity = 'error'
-            report.severity_reason = {
-              :type => Bugsnag::Report::UNHANDLED_EXCEPTION
-            }
-          end
-        end
+      at_exit { $! && at_exit_handler($!) }
+    end
+
+    def at_exit_handler(exception)
+      return if exception.is_a?(SignalException)
+
+      Bugsnag.notify(exception, true) do |report|
+        report.severity = 'error'
+        report.severity_reason = {
+          :type => Bugsnag::Report::UNHANDLED_EXCEPTION
+        }
       end
     end
 
--- a/spec/bugsnag_spec.rb
+++ b/spec/bugsnag_spec.rb
@@ -41,33 +41,58 @@ describe Bugsnag do
       Bugsnag.register_at_exit
     end
 
-    context 'with aliased at_exit' do
-      before do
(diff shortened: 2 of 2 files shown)


bengourley:
Hi @mrnugget, thanks for the PR. At a glance your description of the problem sounds reasonable and the current behaviour undesirable. I'll schedule a proper look at this PR and see where we should take it from here. Cheers!

soulcutter:
im-a-big-fan

snmaynard:
How would we feel about signalexceptions being added to the default ignoreclasses? I'd rather have this behaviour be configurable by users with a sensible default. Right now you can achieve this by adding SignalException to the ignore classes.

Having said that we already have something like this in other integrations like https://github.com/bugsnag/bugsnag-ruby/blob/35b3ef07ef21aab09ad188e0485f1eb6310b07a6/lib/bugsnag/integrations/sidekiq.rb#L25.

I think I'd favour removing all the special cases and ignoring signal exceptions. Thoughts?

mrnugget:
> How would we feel about signalexceptions being added to the default ignoreclasses?

That's probably the cleanest of all solutions! I'd prefer it, yes. 

Also didn't know that there were default ignoreclasses 🙈. A quick search got me to this PR which seems to do the same for other two exceptions in the Sidekiq integration:

https://github.com/bugsnag/bugsnag-ruby/pull/404

So, yes, I strongly agree, let's add `SignalException` to this set 👍

bengourley:
I'm going to close this PR as you can add `SignalException` to the list of `ignore_classes` in your app's configuration to achieve the desired behaviour.

I'm also going to make an internal backlog ticket to add this to the default list of `ignore_classes`, and to remove the specific logic in the sidekiq integration (and any others where it exists).

Thanks all!

PR #483, opened 2018-08-17
Set the args data correctly in DelayedJob plugin when the payload object is a custom class

manojmj92:
## Goal

Example:

We have a custom class that responds to `perform`
```ruby
class NewsLetter
  def initialize(a)
    @a = a
  end

  def perform
  end
end
```
and when we invoke this using `DelayedJob` like
`Delayed::Job.enqueue(NewsLetter.new(2))`,

the `payload` object generated is like 
`#<NewsLetter:0x00007f8cfaa9a860 @a=2>`

Now, this does not respond to the methods `args` or `to_h`, which is what Bugsnag gem currently uses to identify the arguments passed to the job.

Since `args` are not being set in such cases, it becomes very difficult to triage and debug failed `DelayedJob` jobs, because there is no information about the job to begin with:

Eg:

screen shot 2018-08-17 at 12 17 48 pm

As you can see, the only information available about the failed jobs are it's class name and the ID, which is not very helpful in debugging. For debugging to happen effectively, we need the arguments that were passed to this class.

This change is intended to push as much information as possible in the `args` section of the bugsnag payload in such cases, so that debugging becomes easier.

## Tests

Added test cases for where the `DelayedJob` class is a custom class

## Review

  - [x] Final review

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # Bugsnag exception reporter for Ruby
 [![build status](https://travis-ci.org/bugsnag/bugsnag-ruby.svg?branch=master)](https://travis-ci.org/bugsnag/bugsnag-ruby)
+![Gem Version](https://badge.fury.io/rb/bugsnag.svg)
+[![SemVer](https://api.dependabot.com/badges/compatibility_score?dependency-name=bugsnag&package-manager=bundler&version-scheme=semver)](https://dependabot.com/compatibility-score.html?dependency-name=bugsnag&package-manager=bundler&version-scheme=semver)
+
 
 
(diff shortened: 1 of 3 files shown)


Cawllec:
At first glance I don't see any issues with this.  I'll do a full review and look at getting a merge done soon.

manojmj92:
Thanks @Cawllec !

Cawllec:
Hey @manojmj92, since we're rewriting the delayed job plugin [here](https://github.com/bugsnag/bugsnag-ruby/pull/492), we're going to incorporate your change via that PR

PR #491, opened 2018-10-23
Add ruby25 to TravisCI testing

Cawllec:
## Goal
Testing bugsnag against Ruby 2.5

From the PR started [here](https://github.com/bugsnag/bugsnag-ruby/pull/488)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ jobs:
   - stage: test
     env:
     - BUNDLE_VERSION=1.12.0
-    - GEMSETS="test sidekiq coverage"
+    - GEMSETS="test sidekiq"
     rvm: 2.3.0
   - stage: test
     env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,18 +43,23 @@ jobs:
...
   - stage: test
     env:
     - BUNDLE_VERSION=1.12.0
     - GEMSETS="test sidekiq"
     rvm: 2.4.1
+  - stage: test
+    env:
+    - BUNDLE_VERSION=1.12.0
+    - GEMSETS="test sidekiq coverage"
+    rvm: 2.5.1
```
snmaynard: I think we should use 2.5 as we don't want to have to bump for patches. Similar for the other versions.

Cawllec: I've updated to use patch independent versioning in all cases except for `1.9.3`.

PR #492, opened 2018-10-25
Start rewriting delayed job integration

snmaynard:
Beginnings of a delayed job integration rewrite. I am just beginning to get the maze tests written.

The integration should work and may alleviate issues people are having in its current form, but I want to increase test coverage before this gets merged.


lib/bugsnag/integrations/delayed_job.rb
```diff
@@ -54,43 +56,35 @@ def error(job, error)
...
 
-            ::Bugsnag.notify(error, true) do |report|
+            ::Bugsnag.notify(exception, true) do |report|
               report.severity = "error"
               report.severity_reason = {
                 :type => ::Bugsnag::Report::UNHANDLED_EXCEPTION_MIDDLEWARE,
-                :attributes => FRAMEWORK_ATTRIBUTES
+                :attributes => {
+                  :framework => "DelayedJob"
+                }
```
mike-stewart: Is it possible to set the app_type here? Like you do with rake: https://github.com/bugsnag/bugsnag-ruby/blob/56f387ef36c157a2b6fbf7d9e6807d21d1e62393/lib/bugsnag/integrations/rake.rb#L14

Cawllec:
Moved the delayed_job metadata reporting into a middleware, attaching `job` to `request_data` before run to ensure data is attached correctly in both handled and unhandled exceptions.

lib/bugsnag/integrations/delayed_job.rb
```diff
@@ -5,92 +5,34 @@
...
-            ::Bugsnag.notify(error, true) do |report|
+module Delayed
+  module Plugins
+    class Bugsnag < ::Delayed::Plugin
+      callbacks do |lifecycle|
+        lifecycle.around(:invoke_job) do |job, *args, &block|
+          begin
+            ::Bugsnag.configuration.app_type = 'delayed_job'
+            ::Bugsnag.configuration.set_request_data(:delayed_job, job)
+            ::Bugsnag.configuration.internal_middleware.use(::Bugsnag::Middleware::DelayedJob)
```
snmaynard: This should be added once only. I would add it at the top level, on line 2 or something - because at that point we know that they are running delayed job at least.

lib/bugsnag/integrations/delayed_job.rb
```diff
@@ -5,92 +5,34 @@
...
-            end
-
-            ::Bugsnag.notify(error, true) do |report|
+module Delayed
+  module Plugins
+    class Bugsnag < ::Delayed::Plugin
+      callbacks do |lifecycle|
+        lifecycle.around(:invoke_job) do |job, *args, &block|
+          begin
+            ::Bugsnag.configuration.app_type = 'delayed_job'
```
snmaynard: This would ideally be run only once - but I think for now this is ok, until we do the app_type order investigation ticket.

snmaynard:
Can we also base this PR off of the #493 branch? so the diff is easier to read

features/fixtures/delayed_job/app/app/models/test_model.rb
```diff
@@ -0,0 +1,13 @@
+class TestModel < ApplicationRecord
+  def self.print
```
tobyhs: It might be preferable to avoid method names like `print` and `fail` as they override `Kernel#print` and `Kernel#fail`

lib/bugsnag/middleware/delayed_job.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        if job.respond_to?('payload_object')
+          job_data[:active_job] = job.payload_object.job_data if job.payload_object.respond_to?('job_data')
+          job_data[:payload] = construct_job_payload(job.payload_object)
+        end
+
+        if job.respond_to?(:attempts)
+          max_attempts = (job.respond_to?(:max_attempts) && job.max_attempts) || Delayed::Worker.max_attempts
+          job_data[:attempts] = "#{job.attempts + 1} / #{max_attempts}"
+          # +1 as "attempts" is zero-based and does not include the current failed attempt
```
tobyhs: Would it make sense to have the DelayedJob integration/plugin register an after callback on the `:error` event instead? It seems more semantically appropriate, and I think there wouldn't be a need to add one to `job.attempts`.

lib/bugsnag/middleware/delayed_job.rb
```diff
@@ -0,0 +1,71 @@
...
+          :id => job.id
+        }
+        job_data[:priority] = job.priority if job.respond_to?(:priority)
+        job_data[:run_at] = job.run_at if job.respond_to?(:run_at)
+        job_data[:locked_at] = job.locked_at if job.respond_to?(:locked_at)
+        job_data[:locked_by] = job.locked_by if job.respond_to?(:locked_by)
+        job_data[:created_at] = job.created_at if job.respond_to?(:created_at)
+        job_data[:queue] = job.queue if job.respond_to?(:queue)
+
+        if job.respond_to?('payload_object')
```
tobyhs: Use a Symbol for this and the argument to `respond_to?` on the next line (as it seems more conventional to use symbols as arguments to `respond_to?`)

lib/bugsnag/middleware/delayed_job.rb
Cawllec: From doing some investigation the method we're currently using seems to be the most common method of adding error handling to all jobs within an application without disrupting error handling on specific jobs or workers.  However if there is an alternative, it would be good to review it while these changes are being made.

Cawllec:
Due to quick discussion with @kattrali & @duncanhewett a context has been added to the delayed_job middleware.
This is being set to the payloads `method_name` now, but that property won't exist in the event a `job` and `payload` don't exist, leaving the event with no context.
Any input on what context should be by default would be useful.

duncanhewett:
Using class + method name as the context would be better I think.

The default can be left empty and will be defaulted during server-side event processing.

lib/bugsnag/middleware/delayed_job.rb
```diff
@@ -0,0 +1,73 @@
...
+        if job.respond_to?(:payload_object)
+          job_data[:active_job] = job.payload_object.job_data if job.payload_object.respond_to?(:job_data)
+          payload_data = construct_job_payload(job.payload_object)
+          report.context = payload_data[:method_name] if payload_data.include?(:method_name)
+          job_data[:payload] = payload_data
+        end
+
+        if job.respond_to?(:attempts)
+          max_attempts = (job.respond_to?(:max_attempts) && job.max_attempts) || Delayed::Worker.max_attempts
+          job_data[:attempts] = "#{job.attempts + 1} / #{max_attempts}"
```
tobyhs: Would it make more sense to have 2 separate entries for this (an `attempt` and `max_attempts`)? (so the data is easier to deal with, and in case someone wants a custom filter/index on one of these for some reason)

features/support/env.rb
```diff
@@ -18,6 +18,22 @@ def output_logs
...
 end
 
+def install_fixture_gems
+  gem_dir = File.expand_path('../../../', __FILE__)
+  Dir.chdir(gem_dir) do
+    `rm bugsnag-*.gem`
+    `gem build bugsnag.gemspec`
+    Dir.foreach('features/fixtures') do |entry|
+      target_dir = "features/fixtures/#{entry}"
+      if File.directory? (target_dir)
```
tobyhs: Remove the extra space after `File.directory?`

PR #493, opened 2018-10-29
Delayed Job Tests

snmaynard:
Add integration tests for the delayed job integration.

I've added a step to maze to be able to build the current gem and put that into the test docker container. At the moment that invalidates the build because I'm guessing the gem contains the build time - so we need to fix that.

Once we have the framework up and running adding the rests of the tests should be easy.


Cawllec:
I've taken this up, fixed the existing Maze-tests and added a couple of features to the Delayed job tests.  As these tests aren't currently running on CI, reviewers should manually run the `delayed_job` feature.

features/fixtures/delayed_job/Dockerfile
```diff
@@ -1,12 +1,18 @@
 ARG ruby_version
 FROM ruby:$ruby_version
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+WORKDIR /bugsnag
+COPY *.gem ./
+RUN gem unpack bugsnag*.gem /bugsnag
```
snmaynard: Doing this means that you always have to rebuild the docker container I think. Perhaps unpacking outside the container and pulling that into the directory works better?

features/delayed_job.feature
```diff
@@ -0,0 +1,67 @@
+Feature: Bugsnag detects errors in Delayed job workers
+
+Background:
+  Given I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+  And I set environment variable "MAZE_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+  And I configure the bugsnag endpoint
+  And I generate the gem and put it in "delayed_job"
```
snmaynard: Ben has done this in env.rb and I think I prefer that approach. Perhaps we should look at doing the same here.

features/fixtures/delayed_job/app/app/models/test_model.rb
```diff
@@ -0,0 +1,17 @@
+class TestModel < ApplicationRecord
+  def self.print
```
snmaynard: This isnt required

features/fixtures/delayed_job/app/app/models/test_model.rb
```diff
@@ -0,0 +1,17 @@
...
+    puts "yo"
+  end
+
+  def self.fail
```
snmaynard: This isnt required

features/delayed_job.feature
```diff
@@ -0,0 +1,67 @@
...
+
+Scenario: An unhandled RuntimeError sends a report
```
snmaynard: We dont need this test as it just tests a subset of the next test

PR #495, opened 2018-11-13
Fix Sidekiq metadata: Ensures metadata is attached to notifications

Cawllec:
## Goal
Fix the issue where the Sidekiq integration is clearing request data before it can be attached to a notify call.


lib/bugsnag/integrations/sidekiq.rb
```diff
@@ -26,7 +26,7 @@ def call(worker, msg, queue)
         self.class.notify(ex) unless self.class.sidekiq_supports_error_handlers
         raise
       ensure
-        Bugsnag.configuration.clear_request_data
+        Bugsnag.configuration.clear_request_data unless self.class.sidekiq_supports_error_handlers
```
tobyhs: Would you also need to clear request/job data when `sidekiq_supports_error_handlers` is `true` and the job succeeds? There appear to be other places in Sidekiq (besides the main job processing part that invokes server middleware; search for `handle_exception` in the Sidekiq source) that can call registered Sidekiq error handlers and this request/job data shouldn't necessarily leak over.

Cawllec: Yeah, we would.  It looks like individual job callbacks is a pro feature, otherwise all we can subscribe to is `stop` which, if I'm reading this right, only occurs at the end of all jobs currently queued.

I was thinking we could have a flag to see if the `Rescue` was tripped, and avoid removing the data in the `Ensure` if that's the case?

spec/integrations/sidekiq_spec.rb
```diff
@@ -66,4 +66,90 @@ def perform(value)
...
+          chain.add ::Bugsnag::Sidekiq
+        end
+      end
+
+      it "resets with an exception" do
+        expect(Bugsnag.configuration).to receive(:set_request_data).with(:sidekiq, hash_including(:queue => "default"))
+        expect(Bugsnag.configuration).to receive(:clear_request_data).at_least(:once)
+        begin
+          Worker.perform_async(-0)
+          fail("shouldn't be here")
```
tobyhs: It seems more appropriate to have something like `expect { Worker.perform_async(-0) }.to raise_error` instead of this begin/rescue (similar comment for similar lines of code below)

features/sidekiq.feature
```diff
@@ -77,7 +76,6 @@ Scenario Outline: A handled RuntimeError can be notified
   | ruby version | sidekiq_version | created_at_present |
   | 1.9.3        | ~> 2            | false              |
   | 2.0          | ~> 2            | false              |
-  | 2.0          | ~> 3            | true               |
```
snmaynard: Why have these been removed?

features/fixtures/sidekiq/Dockerfile
```diff
@@ -1,6 +1,9 @@
 ARG RUBY_VERSION
 FROM ruby:$RUBY_VERSION
 
+WORKDIR /bugsnag
```
snmaynard: This change should really be applied to all features in a separate PR

lib/bugsnag/integrations/sidekiq.rb
```diff
@@ -24,9 +24,10 @@ def call(worker, msg, queue)
         yield
       rescue Exception => ex
         self.class.notify(ex) unless self.class.sidekiq_supports_error_handlers
+        keep_request_data = self.class.sidekiq_supports_error_handlers
         raise
       ensure
-        Bugsnag.configuration.clear_request_data
+        Bugsnag.configuration.clear_request_data unless keep_request_data
```
snmaynard: We should just use `self.class.sidekiq_supports_error_handlers`. This is not defined in the case where there is no exception currently which is confusing and buggy looking.

snmaynard: we could have a boolean that is `error_raised` that is set to false initially then tru in the rescue. Then also we could have the if statement ammended to check both conditions with a comment explaining why.

Cawllec: Ok, if that makes it more clear

features/fixtures/sidekiq/Dockerfile
Cawllec: This is currently in progress

snmaynard: Where? When will it be done?

features/fixtures/sidekiq/app/Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'bugsnag'
+gem 'bugsnag', path: '/bugsnag'
 gem 'sidekiq', ENV['SIDEKIQ_VERSION']
+gem 'timers', '~> 4.1.1'
```
snmaynard: Why have these gems been added?

features/fixtures/sidekiq/Dockerfile
Cawllec: I'm just running the maze tests, and will PR them once I'm sure they're running correctly

features/fixtures/sidekiq/app/Gemfile
Cawllec: A loading/bundler issue with installing the Gems was occuring with this configuration.  This fixes that, ensuring we can test against the Ruby 2.0/Sidekiq 3 combination

features/sidekiq.feature
Cawllec: Added back in, but added fixed Gems to ensure compatibility.

Cawllec:
Just need the codeclimate allowable complexity to be updated here, I don't have access to that functionality.

snmaynard:
@kattrali can you make the change?

javierjulio:
I see this includes a metadata tab for errors that occur within Sidekiq, but only includes msg and queue. Is there a way for Bugsnag to report the worker args when any error occurs within Sidekiq?

Cawllec:
Hey @javierjulio, the arguments for the job the worker was running with should be present in the `msg => args` section of the sidekiq metadata.  
If that isn't the case, or you aren't seeing that reported, please let us know.

PR #496, opened 2018-11-15
Update maze support

Cawllec:
## Goal

Updates maze tests to reflect changes recently made to Bugsnag-maze-runner.
Includes:

- Removes unnecessary and conflicting steps
- Ensure route tests reflect fixtures
- Add code to remove generate libraries


features/support/env.rb
```diff
@@ -34,14 +33,22 @@ def install_fixture_gems
...
 end
 
-def current_ip
-  if OS.mac?
-    'host.docker.internal'
-  else
-    ip_addr = `ifconfig | grep -Eo 'inet (addr:)?([0-9]*\\\.){3}[0-9]*' | grep -v '127.0.0.1'`
-    ip_list = /((?:[0-9]*\.){3}[0-9]*)/.match(ip_addr)
-    ip_list.captures.first
+def remove_installed_gems
```
snmaynard: Why is this done? Should we use gitignore instead?

features/delayed_job.feature
```diff
@@ -2,7 +2,7 @@ Feature: Bugsnag detects errors in Delayed job workers
 
 Background:
   Given I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
-  And I set environment variable "MAZE_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+  And I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
```
tobyhs: Should this line be removed? It looks like the same step as the previous line. Same question/comment applies to a few other .feature files below.

features/support/env.rb
Cawllec: I didn't like the fact that multiple versions of the Gem could be installed under the `temp-bugsnag-lib` directory in each fixture.  I only noticed when running against 6.9.0 after running against 6.8.0, and had both `bugsnag-6.8.0` and `bugsnag-6.9.0` in the directory, and as there're many fixtures this is a bit of a clean up

snmaynard: Can we add a comment with this clarification?

Cawllec: Added

PR #497, opened 2018-11-19
Mark Bugsnag frames as out of project

Cawllec:
## Goal
To enable easier examination of the Bugsnag reporting process, and why certain events appear as they do. 

## Design

Rather than dropping frames from `lib/bugsnag` they are now included, but will automatically be marked out of project.

## Tests

Unit test added to ensure Bugsnag frames are added correctly and `inProject` is not set.

Will require #496 to be merged first so that Maze tests can be added (in progress)

## Review

  - [ ] Final review


spec/report_spec.rb
```diff
@@ -512,14 +512,15 @@ def gloops
     }
   end
 
-  it "marks the top-most stacktrace line as inProject if necessary" do
+  it "marks the top-most non-bugsnag stacktrace line as inProject if necessary" do
```
snmaynard: This shouldnt be required

spec/report_spec.rb
```diff
@@ -512,14 +512,15 @@ def gloops
...
     Bugsnag.configuration.project_root = File.expand_path File.dirname(__FILE__)
     Bugsnag.notify(BugsnagTestException.new("It crashed"))
```
snmaynard: We should raise this exception so it has a populated stacktrace

spec/report_spec.rb
```diff
@@ -512,14 +512,15 @@ def gloops
...
 
     expect(Bugsnag).to have_sent_notification{ |payload, headers|
       exception = get_exception_from_payload(payload)
       expect(exception["stacktrace"].size).to be >= 1
-      expect(exception["stacktrace"].first["inProject"]).to eq(true)
+      top_frame = get_project_frame(exception["stacktrace"])
```
snmaynard: This should then be able to operate on the stacktrace directly

spec/report_spec.rb
```diff
@@ -1051,11 +1052,11 @@ def gloops
       expect(exception["errorClass"]).to eq("RuntimeError")
       expect(exception["message"]).to eq("'nil' was notified as an exception")
 
-      stacktrace = exception["stacktrace"][0]
-      expect(stacktrace["lineNumber"]).to eq(1047)
+      stacktrace = get_project_frame(exception["stacktrace"])
```
snmaynard: This should be moved into a separate test - as this test is just testing the message. We should probably test that the stacktrace is bugsnag initially then this file after the right number of frames. But that might be covered by what is in the stacktrace spec now?

spec/stacktrace_spec.rb
```diff
@@ -25,6 +26,21 @@
     }
   end
 
+  it "includes bugsnag lines marked out of project" do
```
snmaynard: I dont think this should be here. I think its a test of report and not stacktrace.

spec/stacktrace_spec.rb
```diff
@@ -12,8 +12,9 @@
 
     expect(Bugsnag).to have_sent_notification{ |payload, headers|
       exception = get_exception_from_payload(payload)
-      starting_line = __LINE__ - 10
-      expect(exception["stacktrace"][1]["code"]).to eq({
+      starting_frame = get_project_frame(exception["stacktrace"], 1)
```
snmaynard: Instead of using this function, use a function that explicitly raises and catches an exception to notify - rather than notifying an exception with no stack.

spec/spec_helper.rb
```diff
@@ -84,3 +84,18 @@ def have_sent_notification(&matcher)
     end
   end
 end
+
+def get_project_frame(stacktrace, count=0)
```
snmaynard: I dont think this should be required

lib/bugsnag/stacktrace.rb
```diff
@@ -27,7 +27,7 @@ def initialize(backtrace, configuration)
         # Parse the stacktrace line
 
         # Skip stacktrace lines inside lib/bugsnag
```
tobyhs: I think this comment is no longer applicable

Cawllec:
I've fixed the maze tests that this change has broken.

features/plain_features/report_stack_frames.feature
```diff
@@ -14,8 +14,8 @@ Scenario Outline: Stack frames can be removed
   And the request used the "Ruby Bugsnag Notifier" notifier
   And the request used payload v4 headers
   And the request contained the api key "a35a2a72bd230ac0aa0f52715bbdc6aa"
-  And the "file" of stack frame 0 equals "/usr/src/app/stack_frame_modification/initiators/<initiator>.rb"
```
snmaynard: You should instead update the features to raise the exception so that bugsnag doesnt appear in the stacktrace

Cawllec:
Additional maze changes have been made to ensure stacks from different types of handled errors (a string vs a caught exception) are both tested.

features/steps/ruby_notifier_steps.rb
```diff
@@ -12,4 +12,30 @@
   steps %Q{
     When I set environment variable "#{env_var}" to "#{credentials}@#{current_ip}:#{MOCK_API_PORT}"
   }
+end
+Then(/^the "(.+)" of the top project stackframe equals "(.+)"(?: for request (\d+))?$/) do |element, value, request_index|
+  stacktrace = read_key_path(find_request(request_index)[:body], 'events.0.exceptions.0.stacktrace')
+  frame_index = 0
```
tobyhs: I think you can use `Array#find_index` to make this more concise (probably something like `frame_index = stacktrace.find_index { |frame| frame['file'] !~ %r{lib/bugsnag.*\.rb\z} }`)

features/steps/ruby_notifier_steps.rb
```diff
@@ -12,4 +12,30 @@
   steps %Q{
     When I set environment variable "#{env_var}" to "#{credentials}@#{current_ip}:#{MOCK_API_PORT}"
   }
+end
+Then(/^the "(.+)" of the top project stackframe equals "(.+)"(?: for request (\d+))?$/) do |element, value, request_index|
```
tobyhs: Can we have a blank line between step definitions for organization/easier editor navigation? (between lines 15-16, 28-29)

features/steps/ruby_notifier_steps.rb
```diff
@@ -12,4 +12,30 @@
...
+  stacktrace = read_key_path(find_request(request_index)[:body], 'events.0.exceptions.0.stacktrace')
+  frame_index = 0
+  stacktrace.each do |frame, index|
+    unless /.*lib\/bugsnag.*\.rb/.match(frame["file"])
```
tobyhs: The step definition pattern has "of the top project stackframe", so should this actually check the `inProject` value instead?

Otherwise, if checking against this regular expression is correct, should the step definition pattern have something like "of the top non-Bugsnag stackframe" instead?

features/steps/ruby_notifier_steps.rb
```diff
@@ -12,4 +12,30 @@
...
+      frame_index = index
+      break
+    end
+  end
+  steps %Q{
+    the "#{element}" of stack frame #{frame_index} equals "#{value}"
+  }
+end
+Then(/^the "(.+)" of the top project stackframe equals (\d+)(?: for request (\d+))?$/) do |element, value, request_index|
```
tobyhs: I don't think you need 2 very similar step definitions; you can probably get away with using `(\d+|(".+"))` (so the double quotes are part of the capture for the string case and you can just have `equals #{value}`)

features/plain_features/report_stack_frames.feature
```diff
@@ -14,33 +14,49 @@ Scenario Outline: Stack frames can be removed
...
+  And I wait for 1 second
+  Then I should receive a request
+  And the request used the "Ruby Bugsnag Notifier" notifier
+  And the request used payload v4 headers
+  And the request contained the api key "a35a2a72bd230ac0aa0f52715bbdc6aa"
+  And the "file" of the top non-bugsnag stackframe equals "/usr/src/app/stack_frame_modification/initiators/handled_block.rb"
+  And the "lineNumber" of the top non-bugsnag stackframe equals <lineNumber>
+
+  Examples:
+  | ruby version | lineNumber |
```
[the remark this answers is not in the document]
Cawllec: Removed

features/plain_features/handled_errors.feature
```diff
@@ -67,8 +67,8 @@ Scenario Outline: A handled error can attach metadata in a block
   And the event "severity" equals "warning"
   And the event "severityReason.type" equals "handledException"
   And the exception "errorClass" equals "RuntimeError"
-  And the "file" of stack frame 0 equals "/usr/src/app/handled/block_metadata.rb"
```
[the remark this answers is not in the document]
Cawllec: Removed, and broken the top most scenario out into two separate scenarios for the same reason.

features/plain_features/report_stack_frames.feature
Cawllec: I've updated the scenarios to separate the tests that require this change and those that don't.

PR #499, opened 2018-11-27
Set consistent context in delayed_job integration

mike-stewart:
## Goal

Currently, a context is only set on DelayedJob errors if a display_name is set on the job, which is not always the case. The default context that it falls back to (top stack trace line) doesn't seem useful.

The default results in errors like `ActiveRecord::StatementInvalid app/models/foo.rb:100`. It does not indicate which job is failing, so one cannot understand the problem at glance in the dashboard.

Other integrations attempt to set a useful default context, see:
https://github.com/bugsnag/bugsnag-ruby/blob/56f387ef36c157a2b6fbf7d9e6807d21d1e62393/lib/bugsnag/integrations/resque.rb#L46
https://github.com/bugsnag/bugsnag-ruby/blob/56f387ef36c157a2b6fbf7d9e6807d21d1e62393/lib/bugsnag/middleware/rails3_request.rb#L19

I think it would make sense to use the job name as the default context. This is consistent with other error monitoring tools I've used.

## Tests

Tests pass without changes. Let me know if you think additional coverage is needed.

### Alternative Approaches

I first tried defining a before_notify_callback to override the context, but that did not work.

## Review

  - [x] Final review

@snmaynard @Cawllec

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+* Set default Delayed Job error context to job class
+  | [#499](https://github.com/bugsnag/bugsnag-ruby/pull/499)
+  | [Mike Stewart](https://github.com/mike-stewart)
+
 ## 6.9.0 (12 Nov 2018)
 
 ### Enhancements
--- a/lib/bugsnag/middleware/delayed_job.rb
+++ b/lib/bugsnag/middleware/delayed_job.rb
@@ -24,6 +24,7 @@ module Bugsnag::Middleware
           job_data[:active_job] = job.payload_object.job_data if job.payload_object.respond_to?(:job_data)
           payload_data = construct_job_payload(job.payload_object)
           report.context = payload_data[:display_name] if payload_data.include?(:display_name)
+          report.context ||= payload_data[:class] if payload_data.include?(:class)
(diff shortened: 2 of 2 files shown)


Cawllec:
This LGTM, though I'm going to confirm changing the potential context won't cause any issues first.

Cawllec:
Hi @mike-stewart, I'm going to take it on from here to add some more end-to-end tests.

mike-stewart:
@Cawllec Any ETA on this one?

snmaynard:
We have some longer term work going on adding breadcrumbs to the ruby notifier and we will slot this in after that. Might take a few weeks. You can always pin to your fork until we have got to merging this down if you need it sooner.

mike-stewart:
Any plans to pull this @Cawllec @snmaynard @phillipsam?

phillipsam:
Hi @mike-stewart we are still planning to pull this when priorities allow.

imjoehaines:
Thanks for the PR @mike-stewart! I've fixed the conflict and added an end-to-end test in #615

phillipsam:
A fix for this has now been released in [v6.16.0](https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.16.0).

PR #501, opened 2018-11-30
Add CircularBuffer class and tests

Cawllec:
## Goal
Adds CircularBuffer class and tests with a view to support breadcrumbs.


lib/bugsnag.rb
```diff
@@ -17,6 +17,8 @@
 # as it doesn't auto-configure when loaded
 require "bugsnag/integrations/rack"
 
+require "bugsnag/utility/circular_buffer"
```
tobyhs: In your design doc, I thought you were planning to directly use `Bugsnag::Utility::CircularBuffer` in bugsnag/configuration.rb. If that's the case, you should `require` it in that file instead. However, since you're not using the `CircularBuffer` class yet, I would just not require it until you reach the pull request where you use it.

lib/bugsnag/utility/circular_buffer.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    def max_items=(new_max_items)
+      @max_items = new_max_items
+      trim_buffer
+    end
+
+    private
+    def trim_buffer
+        trim_size = @buffer.size - @max_items
```
tobyhs: Use 2-space indentation

spec/utility/circular_buffer_spec.rb
```diff
@@ -0,0 +1,91 @@
+# encoding: utf-8
+
+require 'spec_helper'
+
+describe Bugsnag::Utility::CircularBuffer do
+  describe "buffer size" do
```
tobyhs: For unit tests, I prefer to have the 2nd level `describe`s use the method names as the doc strings (e.g. this would be `describe '#max_items'`, lines 19-24 would be in its own `describe '#max_items='`)

spec/utility/circular_buffer_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    it "can be changed using max_items =" do
+      buffer = Bugsnag::Utility::CircularBuffer.new(10)
+      buffer.max_items = 17
+
+      expect(buffer.max_items).to equal(17)
+    end
+  end
+
+  describe "buffer items" do
+    it "initializes empty <<" do
```
tobyhs: I don't understand why `<<` is in this describe/doc string

spec/utility/circular_buffer_spec.rb
```diff
@@ -0,0 +1,91 @@
+# encoding: utf-8
+
+require 'spec_helper'
+
+describe Bugsnag::Utility::CircularBuffer do
```
tobyhs: For new spec files, we should probably use `RSpec.describe` for the top-level one (so we can eventually turn off the global DSL described by https://relishapp.com/rspec/rspec-core/v/3-0/docs/configuration/global-namespace-dsl )

spec/utility/circular_buffer_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  describe "buffer items" do
+    it "initializes empty <<" do
+      buffer = Bugsnag::Utility::CircularBuffer.new
+
+      expect(buffer.to_a).to eq([])
+    end
+
+    it "can be added to with <<" do
```
tobyhs: Lines 34-60 should probably be in a `describe '#<<'` block instead (and maybe lines 62-74 in `describe '#each'`, and lines 76-89 in the same `describe '#max_items='` described by a previous comment)

spec/utility/circular_buffer_spec.rb
Cawllec: Personally I prefer tests to be based around expectations of functionality rather than necessarily method testing, as I find `the buffer size can be changed` is more useful when reading the tests than `max_items=`.  However, this is mostly semantics and I don't mind changing the test focus if it helps others read them.

lib/bugsnag.rb
Cawllec: Agreed

spec/utility/circular_buffer_spec.rb
Cawllec: Gotcha, I'll change it over.

spec/utility/circular_buffer_spec.rb
Cawllec: I'll restructure around methods, as I agree it may be preferable.

PR #502, opened 2018-11-30
v6.10.0 Release

Cawllec:
## Goal

Updates Version & Changelog for the v6.10.0 release of: #497 #479 #495


Cawllec:
I'm not sure why codeclimate has only detected this now, it's a pain.  I'll raise a ticket to review codeclimate and possibly create a settings file to ensure it's not going to be creating false positives.

kattrali:
Seems ok to review linting issues on a case-by-case basis. The ones which popped for this PR had similar looking code but weren't related to each other, so I marked them as wontfix.

spec/report_spec.rb
```diff
@@ -1062,7 +1071,7 @@ def gloops
       expect(exception["message"]).to eq("'nil' was notified as an exception")
     }
   end
-
+
```
tobyhs: How did this trailing whitespace get in here?

PR #503, opened 2018-12-04
Breadcrumbs: Add Breadcrumb class

Cawllec:
Adds Bugsnag::Breadcrumbs::Breadcrumb class and tests in order to support Ruby Breadcrumbs feature.

### DIFF
--- a/lib/bugsnag/breadcrumbs/breadcrumb.rb
+++ b/lib/bugsnag/breadcrumbs/breadcrumb.rb
@@ -9,7 +9,7 @@ module Bugsnag::Breadcrumbs
     # @return [Hash, nil] metadata hash containing strings, numbers, or booleans, or nil
     attr_accessor :meta_data
 
-    # @return [Symbol] set to `:auto` if the breadcrumb was automatically generated
+    # @return [Boolean] set to `true` if the breadcrumb was automatically generated
     attr_reader :auto
 
     # @return [Time] a Time object referring to breadcrumb creation time
@@ -36,7 +36,7 @@ module Bugsnag::Breadcrumbs
       @auto = auto == :auto
 
       # Store it as a timestamp for now
-      @timestamp = Time.now
+      @timestamp = Time.now.utc
     end
 
     ##
@@ -69,7 +69,7 @@ module Bugsnag::Breadcrumbs
         :name => @name,
         :type => @type,
         :metaData => @meta_data,
-        :timestamp => @timestamp.utc.iso8601
+        :timestamp => @timestamp.iso8601
       }
     end
   end
--- a/lib/bugsnag/breadcrumbs/breadcrumb.rb
+++ b/lib/bugsnag/breadcrumbs/breadcrumb.rb
@@ -11,7 +11,7 @@ module Bugsnag::Breadcrumbs
       self.message = message
       self.type = type
       self.meta_data = meta_data
-      @auto = auto === :auto
+      @auto = auto == :auto
       @timestamp = Time.now
     end
 
--- a/lib/bugsnag/breadcrumbs/breadcrumb.rb
+++ b/lib/bugsnag/breadcrumbs/breadcrumb.rb
@@ -1,5 +1,4 @@
 module Bugsnag::Breadcrumbs
-
   class Breadcrumb
     attr_accessor :message
     attr_accessor :type
@@ -13,7 +12,7 @@ module Bugsnag::Breadcrumbs
       self.type = type
       self.meta_data = meta_data
       @auto = auto
-      @timestamp = Time.now().utc().strftime('%Y-%m-%dT%H:%M:%S')
+      @timestamp = Time.now.utc.strftime('%Y-%m-%dT%H:%M:%S')
     end
 
     def ignore!
--- a/lib/bugsnag/breadcrumbs/breadcrumb.rb
+++ b/lib/bugsnag/breadcrumbs/breadcrumb.rb
@@ -1,19 +1,34 @@
 module Bugsnag::Breadcrumbs
   class Breadcrumb
-    attr_accessor :message
+    # @return [String] the breadcrumb name
+    attr_accessor :name
+
+    # @return [String] the breadcrumb type
     attr_accessor :type
+
+    # @return [Hash, nil] metadata hash containing strings, numbers, or booleans, or nil
     attr_accessor :meta_data
+
+    # @return [Symbol] set to `:auto` if the breadcrumb was automatically generated
     attr_reader :auto
+
+    # @return [Time] a Time object referring to breadcrumb creation time
     attr_reader :timestamp
 
     ##
-    # Creates a breadcrumb.
+    # Creates a breadcrumb
     #
-    # This will not have been validated, which must occur before this is
-    # attached to a report.
-    def initialize(message, type, meta_data, auto)
+    # This will not have been validated, which must occur before this is attached to a report
+    #
+    # @api private
+    #
+    # @param name [String] the breadcrumb name
+    # @param type [String] the breadcrumb type from Bugsnag::Breadcrumbs::VALID_BREADCRUMB_TYPES
+    # @param meta_data [Hash, nil] a hash containing strings, numbers, or booleans, or nil
+    # @param auto [Symbol] set to `:auto` if the breadcrumb is automatically generated
+    def initialize(name, type, meta_data, auto)
       @should_ignore = false
-      self.message = message
+      self.name = name
       self.type = type
       self.meta_data = meta_data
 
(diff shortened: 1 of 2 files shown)


lib/bugsnag/breadcrumbs/breadcrumb.rb
```diff
@@ -0,0 +1,35 @@
...
+    attr_reader :auto
+    attr_reader :timestamp
+
+    def initialize(message, type, meta_data, auto)
+      @should_ignore = false
+      self.message = message
+      self.type = type
+      self.meta_data = meta_data
+      @auto = auto
+      @timestamp = Time.now.utc.strftime('%Y-%m-%dT%H:%M:%S')
```
snmaynard: Should we store as a time and convert to string later?

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  describe "initialize" do
+    it "should assign arguments correctly" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new("my message", "test type", {:a => 1, :b => 2}, :manual)
+
+      expect(breadcrumb.message).to eq("my message")
+      expect(breadcrumb.type).to eq("test type")
+      expect(breadcrumb.meta_data).to eq({:a => 1, :b => 2})
+      expect(breadcrumb.auto).to eq(:manual)
+
+      expect(breadcrumb.timestamp).to_not be_nil
```
snmaynard: There should be more constraint on this

lib/bugsnag/breadcrumbs/breadcrumb.rb
```diff
@@ -0,0 +1,35 @@
+module Bugsnag::Breadcrumbs
+  class Breadcrumb
+    attr_accessor :message
+    attr_accessor :type
+    attr_accessor :meta_data
+    attr_reader :auto
+    attr_reader :timestamp
+
+    def initialize(message, type, meta_data, auto)
```
snmaynard: yardoc comments?

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,47 @@
+# encoding: utf-8
+
+require 'spec_helper'
+
+require 'bugsnag/breadcrumbs/breadcrumb'
+
+describe Bugsnag::Breadcrumbs::Breadcrumb do
+  describe "initialize" do
+    it "should assign arguments correctly" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new("my message", "test type", {:a => 1, :b => 2}, :manual)
```
snmaynard: is auto not a boolean?

lib/bugsnag/breadcrumbs/breadcrumb.rb
tobyhs: This string should probably end in `Z` to make it conform to a ISO 8601 time

lib/bugsnag/breadcrumbs/breadcrumb.rb
```diff
@@ -0,0 +1,35 @@
...
+      @should_ignore = true
+    end
+
+    def ignore?
+      @should_ignore
+    end
+
+    def to_h
+      {
+        :message => message,
```
tobyhs: I think the error reporting API takes a `name` instead of `message` (from https://bugsnagerrorreportingapi.docs.apiary.io/#reference/0/notify/send-error-reports )

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,47 @@
+# encoding: utf-8
+
+require 'spec_helper'
+
+require 'bugsnag/breadcrumbs/breadcrumb'
+
+describe Bugsnag::Breadcrumbs::Breadcrumb do
+  describe "initialize" do
```
tobyhs: I prefer `describe "#initialize"` (with the `#`) so that the failure message is `Bugsnag::Breadcrumbs::Breadcrumb#initialize should assign arguments correctly` (similarly for the `#to_h` describe block)

spec/breadcrumbs/breadcrumb_spec.rb
Cawllec: Yes, it will be a boolean inside the class, but not the leave_breadcrumbs method. I'll update that.

lib/bugsnag/breadcrumbs/breadcrumb.rb
tobyhs: If we're starting to adopt yardoc in bugsnag-ruby, there should be `@param`, `@return`, etc. comments (see https://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md , but don't follow that first snippet which actually RDoc instead). Also, `#initialize` should probably have an `# @api private` as I don't think end users will need to create a `Breadcrumb`.

lib/bugsnag/breadcrumbs/breadcrumb.rb
```diff
@@ -0,0 +1,57 @@
...
+    # Outputs the breadcrumb data in a formatted hash.
+    #
+    # These adhere to the breadcrumb format as defined in the Bugsnag error 
+    # reporting API
+    def to_h
+      {
+        :name => @message,
+        :type => @type,
+        :metaData => @meta_data,
+        :timestamp => @timestamp.utc.strftime('%Y-%m-%dT%H:%M:%SZ')
```
tobyhs: As mentioned in https://github.com/bugsnag/bugsnag-ruby/pull/503#discussion_r239111307 you can use the `iso8601` method, but you'll probably need to `require 'time'` in this file for that to work.

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      expect(breadcrumb.auto).to eq(false)
+    end
+
+    it "is true if auto argument == :auto" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new(nil, nil, nil, :auto)
+
+      expect(breadcrumb.auto).to eq(true)
+    end
+
+    it "if false if auto argument is anything else" do
```
tobyhs: Replace "if" with "is"

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      expect(breadcrumb.auto).to eq(false)
+    end
+  end
+
+  describe "#timestamp" do
+    it "is stored as a timestamp" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new(nil, nil, nil, nil)
+
+      expect(breadcrumb.timestamp).to be_within(0.5.second).of Time.now
```
tobyhs: Avoid the `second` method unless you're willing to bring in activesupport as an explicit development dependency

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+  end
+
+  describe "#ignore" do
```
tobyhs: This should be `#ignore?` (to match the actual method name called by all examples under this describe block)

lib/bugsnag/breadcrumbs/breadcrumb.rb
```diff
@@ -0,0 +1,76 @@
...
+    # @return [String] the breadcrumb name
+    attr_accessor :name
+
+    # @return [String] the breadcrumb type
+    attr_accessor :type
+
+    # @return [Hash, nil] metadata hash containing strings, numbers, or booleans, or nil
+    attr_accessor :meta_data
+
+    # @return [Symbol] set to `:auto` if the breadcrumb was automatically generated
```
snmaynard: boolean

lib/bugsnag/breadcrumbs/breadcrumb.rb
```diff
@@ -0,0 +1,76 @@
...
+      @should_ignore = false
+      self.name = name
+      self.type = type
+      self.meta_data = meta_data
+
+      # Use the symbol comparison to improve readability of breadcrumb creation
+      @auto = auto == :auto
+
+      # Store it as a timestamp for now
+      @timestamp = Time.now
```
snmaynard: Use UTC

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    it "is assigned in #initialize" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new("my message", nil, nil, nil)
+
+      expect(breadcrumb.name).to eq("my message")
+    end
+  end
+
+  describe "#type" do
+    it "is assigned in #initialize" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new(nil, "test type", nil, nil)
```
snmaynard: its weird that you can set this to an invalid type to me.

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      breadcrumb.ignore!
+
+      expect(breadcrumb.ignore?).to eq(true)
+    end
+  end
+
+  describe "#to_h" do
+    it "outputs as a hash" do
+      breadcrumb = Bugsnag::Breadcrumbs::Breadcrumb.new("my message", "test type", {:a => 1, :b => 2}, :manual)
+      output = breadcrumb.to_h
```
snmaynard: Check there are no extra keys too

lib/bugsnag/breadcrumbs/breadcrumb.rb
Cawllec: We convert to utc before outputting it in the hash

snmaynard: Do it here so you dont have different timezones in the library. Its a best practice to convert to UTC immediately.

spec/breadcrumbs/breadcrumb_spec.rb
Cawllec: As this may be overwritten by callbacks anyway, it seems better to check this in a single place (the validator) than in the breadcrumb.

snmaynard: in isolation i would have the setter validate. Its complicated by the metadata validation where that isnt possible however.

spec/breadcrumbs/breadcrumb_spec.rb
Cawllec: Doing so with better a better defined hash check.

Cawllec: [here](https://github.com/bugsnag/bugsnag-ruby/blob/8e2042e32ceddc34295011ebd244bf05a53a54e0/spec/breadcrumbs/breadcrumb_spec.rb#L82)

spec/breadcrumbs/breadcrumb_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      timestamp_regex = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z$/
+
+      expect(output).to match(
+        :name => "my message",
+        :type => "test type",
+        :metaData => {
+          :a => 1,
+          :b => 2
+        },
+        :timestamp => match(timestamp_regex)
```
snmaynard: Here you should expect it to match timestamp.iso8601 rather than just regexing imo

PR #505, opened 2018-12-05
Breadcrumbs: Add Breadcrumb validator

Cawllec:
## Goal
Adds Bugsnag::Breadcrumbs::Validator class and tests in order to support Ruby Breadcrumbs feature.


lib/bugsnag/breadcrumbs/validator.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      # Check type is valid, set to manual otherwise
+      unless VALID_BREADCRUMB_TYPES.include?(breadcrumb.type)
+        @configuration.warn("Invalid type: #{breadcrumb.type} for breadcrumb: #{breadcrumb.message}, defaulting to #{MANUAL_BREADCRUMB_TYPE}")
+        breadcrumb.type = MANUAL_BREADCRUMB_TYPE
+      end
+
+      # If auto is true, check type is in automatic_breadcrumb_types
+      if breadcrumb.auto && !@configuration.automatic_breadcrumb_types.include?(breadcrumb.type)
```
Cawllec: There's a rubocop issue generated here, about using a guard clause, but I'm not sure how that's expected to work in this circumstance.

tobyhs: I think it wants the code to look something like:
```ruby
      return unless breadcrumb.auto && !@configuration.automatic_breadcrumb_types.include?(breadcrumb.type)

      @configuration.warn("Automatic breadcrumb of type #{breadcrumb.type} ignored: #{breadcrumb.message}")
      breadcrumb.ignore!
```
but that looks horrible in this case.

spec/breadcrumbs/validator_spec.rb
```diff
@@ -0,0 +1,254 @@
+# encoding: utf-8
+require 'spec_helper'
+
+require 'bugsnag/breadcrumbs/types'
+require 'bugsnag/breadcrumbs/validator'
+
+RSpec.describe Bugsnag::Breadcrumbs::Validator do
+  before do
+    @automatic_breadcrumb_types = [
```
tobyhs: I prefer using `let` ( http://www.betterspecs.org/#let ) for these instead of instance variables, as it tends to make it easier to refactor tests in case local variables are more appropriate, and to override values in certain situations.

spec/breadcrumbs/validator_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+      Bugsnag::Breadcrumbs::STATE_BREADCRUMB_TYPE
+    ]
+
+    @breadcrumb_auto = false
+    @breadcrumb_message = "Valid message"
+    @breadcrumb_type = Bugsnag::Breadcrumbs::ERROR_BREADCRUMB_TYPE
+    @breadcrumb_meta_data = {}
+  end
+
+  describe "#Validate" do
```
tobyhs: lower case v

spec/breadcrumbs/validator_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+    it "does not 'ignore!' a valid breadcrumb" do
+      config = double
```
tobyhs: It would be preferable to use `instance_double` instead of `double` as `instance_double` verifies that you don't stub something in an incorrect way. However, it would depend on the `Configuration` and `Breadcrumb` class changes first, so it might be good to organize the pull requests in a way that you can use `instance_double` here.

lib/bugsnag/breadcrumbs/validator.rb
```diff
@@ -0,0 +1,47 @@
...
+      @configuration = configuration
+    end
+
+    ##
+    # Validates a given breadcrumb.
+    def validate(breadcrumb)
+      # Check message length
+      if breadcrumb.message.size > MAX_MESSAGE_LENGTH
+        @configuration.warn("Breadcrumb message trimmed to length #{MAX_MESSAGE_LENGTH}.  Original message: #{breadcrumb.message}")
+        breadcrumb.message.slice!(MAX_MESSAGE_LENGTH...breadcrumb.message.size)
```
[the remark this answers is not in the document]
Cawllec: If it's passed to the breadcrumb initializer it should be a separate value, correct?  So even if it's mutated here, it won't effect anything upstream.

lib/bugsnag/breadcrumbs/validator.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      # Check meta_data hash doesn't contain complex values
+      unless valid_meta_data_types?(breadcrumb.meta_data)
+        @configuration.warn("Breadcrumb #{breadcrumb.message} meta_data has values other than strings, numbers, or booleans, dropping: #{breadcrumb.meta_data}")
+        breadcrumb.meta_data = {}
```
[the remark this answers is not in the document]
Cawllec: This was the original intent of the spec, I just read it wrong

lib/bugsnag/breadcrumbs/validator.rb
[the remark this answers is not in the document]
snmaynard: Strings in ruby are mutable and passedin by reference. So as you currently have it it will mutate the argument passed in.

```ruby
a = "hello there this string is over 30 characters long as an example"
Bugsnag.leave_breadcrumb(a)
puts a
```

Will print "hello there this string is ove"

spec/breadcrumbs/validator_spec.rb
Cawllec: I've made some progress towards this, although I still need to do a bunch of refactoring.  I'm happy for this PR to sit here for now while I make the configuration changes.

spec/breadcrumbs/validator_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+
+  describe "#validate" do
+    it "does not 'ignore!' a valid breadcrumb" do
+      config = instance_double("Configuration")
+      allow(config).to receive(:automatic_breadcrumb_types).and_return(automatic_breadcrumb_types)
+      validator = Bugsnag::Breadcrumbs::Validator.new(config)
+
+      breadcrumb = instance_double("Breadcrumb")
+
+      allow(breadcrumb).to receive(:meta_data=)
```
tobyhs: You can probably make these a little bit more concise by placing the method/message stubs with the initialization:
```ruby
breadcrumb = instance_double(
  Bugsnag::Breadcrumbs::Breadcrumb,
  :meta_data= => meta_data,
  :auto => auto,
  :name => name,
  :type => type,
  :meta_data => meta_data,
)
```

spec/breadcrumbs/validator_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+
+        allow(breadcrumb).to receive_messages({
+          :auto => auto,
+          :name => name,
+          :type => type,
+          :meta_data => meta_data
+        })
+
+        expect(breadcrumb).to_not receive(:ignore!)
+        expected_string_1 = "Breadcrumb #{breadcrumb.name} meta_data #{:array}:#{meta_data[:array]} has been dropped for having an invalid data type"
```
tobyhs: Relace the `#{:array}` with just `array` (similarly for the next 2 lines)

lib/bugsnag/breadcrumbs/validator.rb
```diff
@@ -0,0 +1,53 @@
+require 'bugsnag/breadcrumbs/breadcrumbs'
+
+module Bugsnag::Breadcrumbs
+  class Validator
```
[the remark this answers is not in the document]
Cawllec: Added, although it's only appearing in file changes

PR #506, opened 2018-12-06
Added yard docs to circular buffer class.

Cawllec:
## Goal

Adds Yard doc comments to the circular buffer class.

### DIFF
--- a/lib/bugsnag/utility/circular_buffer.rb
+++ b/lib/bugsnag/utility/circular_buffer.rb
@@ -2,23 +2,48 @@ module Bugsnag::Utility
   class CircularBuffer
     include Enumerable
 
+    # @return [Numeric] the current maximum allowable number of items
     attr_reader :max_items
 
+    ##
+    # Creates a circular buffer
+    #
+    # @api private
+    #
+    # @param max_items [Integer] the initial maximum number of items
     def initialize(max_items = 25)
       @max_items = max_items
       @buffer = []
     end
 
+    ##
+    # Adds an item to the circular buffer
+    #
+    # If this causes the buffer to exceed its maximum items, the oldest item will be removed
+    #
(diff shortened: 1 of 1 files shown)


lib/bugsnag/utility/circular_buffer.rb
```diff
@@ -2,30 +2,57 @@ module Bugsnag::Utility
...
       @max_items = max_items
       @buffer = []
     end
 
+    ##
+    # Adds an item to the circular buffer
+    #
+    # If this causes the buffer to exceed its maximum items, the oldest item will be removed
+    #
+    # @param item [Any] the item to add to the buffer
```
tobyhs: This should probably be `[Object]` (although I could be wrong and wouldn't be surprised if I wasn't aware of some `[Any]` or `[any]` convention in yardoc)

lib/bugsnag/utility/circular_buffer.rb
```diff
@@ -2,30 +2,57 @@ module Bugsnag::Utility
   class CircularBuffer
     include Enumerable
 
+    # @return [Numeric] the current maximum allowable number of items
     attr_reader :max_items
 
+    ##
+    # Creates a circular buffer
```
tobyhs: The main description portion of yard comments for `#initialize` is not needed as yard is smart enough to generate a generically useful one (but leave the `@param` comment in)

lib/bugsnag/utility/circular_buffer.rb
```diff
@@ -2,30 +2,57 @@ module Bugsnag::Utility
...
+    #
+    # @api private
```
tobyhs: If users aren't exposed to the `CircularBuffer` in any direct way, move this `@api private` comment to the class (right before line 2). Also related, add a brief comment comment for the `CircularBuffer` class.

PR #507, opened 2018-12-06
Breadcrumbs: Configuration updates

Cawllec:
## Goal

Update the Configuration class in order to support breadcrumbs.

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -38,8 +38,14 @@ module Bugsnag
     attr_accessor :auto_capture_sessions
     attr_accessor :track_sessions
     attr_accessor :session_endpoint
+
+    ##
+    # @param [Array] array of strings indicating allowable automatic breadcrumb types
+    # @return [Array] indicates the allowable automatic breadcrumbs
     attr_accessor :automatic_breadcrumb_types
 
+    ##
+    # @return [Integer] the maximum allowable amount of breadcrumbs per thread
     attr_reader :max_breadcrumbs
 
     API_KEY_REGEX = /[0-9a-f]{32}/i
@@ -205,6 +211,8 @@ module Bugsnag
 
     ##
     # Sets the maximum allowable amount of breadcrumbs
+    #
+    # @param [Integer] the new maximum breadcrumb limit
     def max_breadcrumbs=(new_max_breadcrumbs)
       @max_breadcrumbs = new_max_breadcrumbs
       breadcrumbs.max_items = new_max_breadcrumbs
@@ -212,12 +220,16 @@ module Bugsnag
 
     ##
     # Returns the breadcrumb circular buffer
+    #
+    # @return [Bugsnag::Utility::CircularBuffer] a thread based circular buffer containing breadcrumbs
     def breadcrumbs
       request_data[:breadcrumbs] ||= Bugsnag::Utility::CircularBuffer.new(@max_breadcrumbs)
     end
 
     ##
     # Stores callbacks that will be run before a breadcrumb is left
+    #
+    # @return [Array] a thread independent array of callback functions
     def before_breadcrumb_callbacks
       @before_breadcrumb_callbacks ||= []
     end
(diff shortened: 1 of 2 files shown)


Cawllec:
This is a first iteration, without tests as of yet, so any reviews should be very quick.

lib/bugsnag/configuration.rb
```diff
@@ -36,6 +37,9 @@ class Configuration
     attr_accessor :auto_capture_sessions
     attr_accessor :track_sessions
     attr_accessor :session_endpoint
+    attr_accessor :automatic_breadcrumb_types
+
+    attr_reader :max_breadcrumbs
```
tobyhs: If you're starting to adopt yardoc, it would be good to add documentation/comments to new methods that you add/create.

Cawllec: To keep things consistent, I was going to add Yardoc comments to new files, and keep existing files as they are, and raise a ticket to Yardoc the whole repo.

spec/configuration_spec.rb
```diff
@@ -222,4 +222,92 @@ def debug(name, &block)
...
+    it "returns a different buffer on different threads" do
+      buffer = subject.breadcrumbs
+      buffer << 1
+
+      second_buffer = nil
+      Thread.new { second_buffer = subject.breadcrumbs; second_buffer << 2 }.join
+
+      expect(buffer.to_a).to eq([1])
+      expect(second_buffer.to_a).to eq([2])
+      expect(buffer).to_not eq(second_buffer)
```
snmaynard: You don't need this expect

lib/bugsnag/configuration.rb
Cawllec: For the sake of these actually getting written, i've added the yardoc comments now and will continue to in all the other changes.

spec/configuration_spec.rb
```diff
@@ -222,4 +222,92 @@ def debug(name, &block)
...
+      expect(buffer.max_items).to eq(Bugsnag::Configuration::DEFAULT_MAX_BREADCRUMBS)
+
+      subject.max_breadcrumbs = 5
+
+      expect(buffer.max_items).to eq(5)
+    end
+  end
+
+  describe "#automatic_breadcrumb_types" do
+    it "defaults to Bugsnag::Breadcrumbs::VALID_BREADCRUMB_TYPES" do
```
snmaynard: Having some tests around being able to change this would be good

spec/configuration_spec.rb
```diff
@@ -222,4 +222,92 @@ def debug(name, &block)
...
+    end
+
+    it "stays the same across threads" do
+      first_array = subject.before_breadcrumb_callbacks
+      first_array << 1
+
+      second_array = nil
+      Thread.new { second_array = subject.before_breadcrumb_callbacks; second_array << 2}.join
+
+      expect(first_array).to eql(second_array)
```
snmaynard: I'd drop this expectation

lib/bugsnag/configuration.rb
```diff
@@ -68,6 +75,12 @@ def initialize
       self.notify_release_stages = nil
       self.auto_capture_sessions = false
       self.session_endpoint = DEFAULT_SESSION_ENDPOINT
+      # All valid breadcrumb types should be allowable initially
+      self.automatic_breadcrumb_types = Bugsnag::Breadcrumbs::VALID_BREADCRUMB_TYPES.clone
```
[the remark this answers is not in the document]
Cawllec: Yes - and I'm adding the extra test to verify it is mutable.

lib/bugsnag/configuration.rb
```diff
@@ -37,6 +39,18 @@ class Configuration
     attr_accessor :track_sessions
     attr_accessor :session_endpoint
 
+    ##
+    # @return [Array] array of strings indicating allowable automatic breadcrumb types
```
snmaynard: Should do [Array<String>] and similar for others

PR #508, opened 2018-12-08
Set the rails logger after initialisation

iainbeeston:
## Goal

Currently, bugsnag logging does not work in rails unless you explicitly set the logger in your bugsnag initializer, and so all logging messages are being lost. I'd like to fix that so that it works out of the box.

### Changed

I've moved the code to set up the bugsnag logger to run after the railtie has been initialized. This fixes the issue for me (I believe because the rails logger has not been set up yet in the `before_initialize` hook, but is available in `after_initialize`).

## Tests

I think there are no existing tests for the railtie?

## Review

  - [ ] Final review

### DIFF
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -22,7 +22,6 @@ module Bugsnag
       # Skipping API key validation as the key may be set later in an
       # initializer. If not, the key will be validated in after_initialize.
       Bugsnag.configure(false) do |config|
-        config.logger = ::Rails.logger
         config.release_stage = ENV["BUGSNAG_RELEASE_STAGE"] || ::Rails.env.to_s
         config.project_root = ::Rails.root.to_s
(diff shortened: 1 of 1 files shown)


lib/bugsnag/integrations/railtie.rb
```diff
@@ -22,7 +22,6 @@ class Railtie < Rails::Railtie
       # Skipping API key validation as the key may be set later in an
       # initializer. If not, the key will be validated in after_initialize.
       Bugsnag.configure(false) do |config|
-        config.logger = ::Rails.logger
```
tobyhs: I believe the logger was set here so that any log output from Bugsnag that occurred in a config/initializers/ file would end up in the Rails log.

When trying the example at https://github.com/bugsnag/bugsnag-ruby/tree/800ea83ee4885df2e2f2c48570e13ffa7e75b9b8/example/rails-51 with your change, Bugsnag log output resulting from a file in config/initializers/ doesn't end up in log/development.log.

tomlongridge:
Closing, given the limitations noted above.

PR #511, opened 2018-12-11
Breadcrumbs/leave breadcrumb

Cawllec:
## Goal

Adds the `leave_breadcrumb` method and associated tests for the Breadcrumbs implementation.

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -29,6 +29,10 @@ require "bugsnag/middleware/callbacks"
 require "bugsnag/middleware/classify_error"
 require "bugsnag/middleware/delayed_job"
 
+require "bugsnag/breadcrumbs/validator"
+require "bugsnag/breadcrumbs/breadcrumb"
+require "bugsnag/breadcrumbs/breadcrumbs"
+
 module Bugsnag
   LOCK = Mutex.new
   INTEGRATIONS = [:resque, :sidekiq, :mailman, :delayed_job, :shoryuken, :que]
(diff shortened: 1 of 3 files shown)


spec/bugsnag_spec.rb
```diff
@@ -133,4 +133,153 @@ module Kernel
...
+        :type => Bugsnag::Breadcrumbs::MANUAL_BREADCRUMB_TYPE,
+        :metaData => {
+          :a => 1
+        },
+        :timestamp => match(timestamp_regex)
+      })
+    end
+
+    it "runs callbacks before leaving" do
+      Bugsnag.configuration.before_breadcrumb_callbacks << Proc.new { |breadcrumb|
```
tobyhs: I normally see `do` + `end` for multi-line blocks instead of curly braces ("Avoid using {...} for multi-line blocks" at https://github.com/rubocop-hq/ruby-style-guide#single-line-blocks )

lib/bugsnag.rb
```diff
@@ -189,6 +193,36 @@ def load_integration(integration)
...
+
+      # Skip if it's already invalid
+      unless breadcrumb.ignore?
+        # Run callbacks
+        configuration.before_breadcrumb_callbacks.each do |c|
+          (c.arity > 0 ? c.call(breadcrumb) : c.call)
+          break if breadcrumb.ignore?
+        end
+
+        # Validate again incase of callback alteration
```
tobyhs: "in case" (with a space in between)

PR #512, opened 2018-12-11
Breadcrumbs: Update reports class, add Breadcrumbs middleware

Cawllec:
## Goal

Adds breadcrumbs middleware and enables the report to receive them.

TODO: Tests

### DIFF
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -168,7 +168,7 @@ module Bugsnag
 
     ##
     # Generates a summary to be attached as a breadcrumb
-    def get_summary
+    def summary
       {
         :name => name,
         :message => message,
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -23,13 +23,16 @@ module Bugsnag
     attr_accessor :api_key
     attr_accessor :app_type
     attr_accessor :app_version
+    attr_accessor :breadcrumbs
     attr_accessor :configuration
     attr_accessor :context
     attr_accessor :delivery_method
     attr_accessor :exceptions
     attr_accessor :hostname
     attr_accessor :grouping_hash
+    attr_accessor :message
     attr_accessor :meta_data
+    attr_accessor :name
     attr_accessor :raw_exceptions
     attr_accessor :release_stage
     attr_accessor :session
@@ -51,9 +54,12 @@ module Bugsnag
       self.api_key = configuration.api_key
       self.app_type = configuration.app_type
       self.app_version = configuration.app_version
+      self.breadcrumbs = {}
       self.delivery_method = configuration.delivery_method
       self.hostname = configuration.hostname
+      self.message = exception.message
       self.meta_data = {}
+      self.name = exceptions.class
       self.release_stage = configuration.release_stage
       self.severity = auto_notify ? "error" : "warning"
       self.severity_reason = auto_notify ? {:type => UNHANDLED_EXCEPTION} : {:type => HANDLED_EXCEPTION}
@@ -110,7 +116,14 @@ module Bugsnag
       payload_event = Bugsnag::Cleaner.clean_object_encoding(payload_event)
 
       # filter out sensitive values in (and cleanup encodings) metaData
-      payload_event[:metaData] = Bugsnag::Cleaner.new(configuration.meta_data_filters).clean_object(meta_data)
+      filter_cleaner = Bugsnag::Cleaner.new(configuration.meta_data_filters)
+      payload_event[:metaData] = filter_cleaner.clean_object(meta_data)
+      payload_event[:breadcrumbs] = breadcrumbs.map do |raw_crumb|
+        breadcrumb = raw_crumb.to_h
+        filter_cleaner.clean_object(breadcrumb[:metaData])
+        breadcrumb
+      end
+
       payload_event.reject! {|k,v| v.nil? }
 
       # return the payload hash
@@ -151,6 +164,16 @@ module Bugsnag
     # Tells the client this report should not be sent.
     def ignore!
       @should_ignore = true
+    end
+
+    ##
+    # Generates a summary to be attached as a breadcrumb
+    def get_summary
+      {
+        :name => name,
+        :message => message,
+        :severity => severity
+      }
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/bugsnag/report.rb
```diff
@@ -110,7 +116,14 @@ def as_json
       payload_event = Bugsnag::Cleaner.clean_object_encoding(payload_event)
 
       # filter out sensitive values in (and cleanup encodings) metaData
-      payload_event[:metaData] = Bugsnag::Cleaner.new(configuration.meta_data_filters).clean_object(meta_data)
+      filter_cleaner = Bugsnag::Cleaner.new(configuration.meta_data_filters)
+      payload_event[:metaData] = filter_cleaner.clean_object(meta_data)
+      payload_event[:breadcrumbs] = breadcrumbs.map do |raw_crumb|
+        breadcrumb = raw_crumb.to_h
```
tobyhs: if `breadcrumbs` is an `Array` of `Breadcrumb` objects, should one of these be `breadcrumb` (instead of `raw_crumb`)? It seems like the variable names should be reversed, as raw implies something more primitive.

Alternatively, the block variable can be `breadcrumb` but the converted one can be named something like `breadcrumb_hash`.

lib/bugsnag/report.rb
```diff
@@ -153,6 +166,16 @@ def ignore!
       @should_ignore = true
     end
 
+    ##
+    # Generates a summary to be attached as a breadcrumb
```
tobyhs: Add an `@return` comment for this method

spec/report_spec.rb
```diff
@@ -1092,6 +1092,104 @@ def gloops
     }
   end
 
+  describe "breadcrumbs" do
+    let(:timestamp_regex) { /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z$/ }
+
+    it "includes left breadcrumbs" do
+      Bugsnag.leave_breadcrumb("Test breadcrumb")
+      notify_test_exception
+      expect(Bugsnag).to have_sent_notification{ |payload, headers|
```
tobyhs: Use `do` + `end`, although I suspect there could be some Ruby precedence issue that prevents this from working; if that's the case, add a space before the `{` (same comment applies to a few other lines below)

spec/report_spec.rb
```diff
@@ -1092,6 +1092,104 @@ def gloops
...
+  describe "#summary" do
+    it "provides a hash of the name, message, and severity" do
+      begin
+        1/0
+      rescue ZeroDivisionError => e
+        report = Bugsnag::Report.new(e, Bugsnag.configuration)
+        expect(report.name).to eq("ZeroDivisionError")
+        expect(report.message).to eq("divided by 0")
+
+        expect(report.summary).to match({
```
tobyhs: This should probably use `eq` instead of `match` (same with line 1185)

lib/bugsnag/report.rb
```diff
@@ -51,9 +54,12 @@ def initialize(exception, passed_configuration, auto_notify=false)
       self.api_key = configuration.api_key
       self.app_type = configuration.app_type
       self.app_version = configuration.app_version
+      self.breadcrumbs = []
       self.delivery_method = configuration.delivery_method
       self.hostname = configuration.hostname
+      self.message = defined?(exception.message) ? exception.message : exception.to_s
       self.meta_data = {}
+      self.name = exception.class.to_s
```
tobyhs: If `exception` is a string (which is likely true if `exception` doesn't respond to `message`) instead of an exception, would it make more sense to default to `RuntimeError` instead (because code later on will convert it to that)?

Cawllec: I think the breadcrumb name should reflect what was notified explicitly.  In this case the result would be `String` which on a breadcrumb with limited information I think is more telling than `RuntimeError` would be.

snmaynard: Users will use this information to try and find the error in the UI so switching between class and message will be confusing.

Cawllec: That's fair, I hadn't considered it in that context.

lib/bugsnag/report.rb
```diff
@@ -23,13 +23,16 @@ class Report
...
+    attr_accessor :breadcrumbs
     attr_accessor :configuration
     attr_accessor :context
     attr_accessor :delivery_method
     attr_accessor :exceptions
     attr_accessor :hostname
     attr_accessor :grouping_hash
+    attr_accessor :message
     attr_accessor :meta_data
+    attr_accessor :name
```
tobyhs: After digging through the existing code some more, it looks like it would be easier to define a plain `name` method (instead of `attr_accessor :name`) that returns `exceptions[0][:errorClass]` as that would be consistent (because the `exceptions` attribute also makes the exception go through the `#error_class` method to find its error class or "name", which your code further below doesn't do now).

lib/bugsnag/report.rb
```diff
@@ -23,13 +23,16 @@ class Report
...
     attr_accessor :app_type
     attr_accessor :app_version
+    attr_accessor :breadcrumbs
     attr_accessor :configuration
     attr_accessor :context
     attr_accessor :delivery_method
     attr_accessor :exceptions
     attr_accessor :hostname
     attr_accessor :grouping_hash
+    attr_accessor :message
```
tobyhs: Similar to the comment a few lines after this, I think you can remove this and change your `#summary` method to use `exceptions[0][:message]`

lib/bugsnag/report.rb
Cawllec: Keeping the `message` and `name` as it is now allows the user to override how the follow-up breadcrumb will be displayed.  If they want to assign a custom name/message that will allow them to navigate breadcrumbs more easily they should be able to.

lib/bugsnag/report.rb
```diff
@@ -153,6 +162,26 @@ def ignore!
       @should_ignore = true
     end
 
+    ##
+    # Generates a summary to be attached as a breadcrumb
+    #
+    # @return [Hash] a Hash containing the report's error class, error message, and severity
+    def summary
+      # Guard against the exceptions array being removed/changed or emptied here
+      if !exceptions.is_a?(Array) || exceptions.first.nil?
```
tobyhs: I think this would be easier to read if you changed the condition to `exceptions.respond_to?(:first) && exceptions.first` and put the known error class case first.

PR #516, opened 2018-12-14
Breadcrumbs: Add Mongo integration

Cawllec:
## Goal

Add Mongo integration to support capturing breadcrumbs


features/rails_features/breadcrumbs.feature
```diff
@@ -56,3 +56,24 @@ Scenario Outline: SQL Breadcrumb
...
+  Given I set environment variable "RUBY_VERSION" to "<ruby_version>"
+  And I start the service "rails<rails_version>"
+  And I wait for the app to respond on port "6128<rails_version>"
+  When I navigate to the route "/mongo/crash" on port "6128<rails_version>"
+  Then I should receive a request
+  And the request is a valid for the error reporting API
+  And the request used the "Ruby Bugsnag Notifier" notifier
+  And the request contained the api key "a35a2a72bd230ac0aa0f52715bbdc6aa"
+  And the event has a "process" breadcrumb named "Mongo query started"
+  And the event has a "process" breadcrumb named "Mongo query succeeded"
```
tobyhs: Can you add a test for the failed event from a separate route/controller-action? (I think causing a duplicate key error on a collection with a unique index should trigger a failed event if you can't think of some other way)

features/rails_features/breadcrumbs.feature
```diff
@@ -56,3 +56,24 @@ Scenario Outline: SQL Breadcrumb
     | 2.4          | 5             |
     | 2.5          | 3             |
     | 2.5          | 5             |
+
+Scenario Outline: Mongo breadcrumbs
```
tobyhs: I would put this in a separate feature file as it isn't related to the other activesupport/Rails related breadcrumbs.

PR #518, opened 2019-01-02
Remove 404'ing link

tibbon:
Remove a 404'ing link from the Readme


Cawllec:
Thanks for the PR @tibbon, but I think we'll redirect it to https://github.com/bugsnag/bugsnag-ruby/tree/master/example instead (as that's where the examples moved to).  If you want to make that change here, I'll merge that down :).

Cawllec:
Closing this one in favour of #533

PR #519, opened 2019-01-03
Breadcrumbs/Mongo: Remove 'started' breadcrumb, add collection detail

Cawllec:
## Goal

Removes the "started" breadcrumb, while adding the commands collection name.

Adds filter keys in the format: `{"filter_1":"?","filter_2":"?"}` etc.

Adds new and improves existing end-to-end tests for mongo breadcrumbs.

### DIFF
--- a/lib/bugsnag/integrations/mongo.rb
+++ b/lib/bugsnag/integrations/mongo.rb
@@ -56,8 +56,8 @@ module Bugsnag
         collection_key = event.command_name == "getMore" ? "collection" : event.command_name
         meta_data[:collection] = command[collection_key]
         unless command["filter"].nil?
-          filters = command["filter"].map { |key, _v| [key, '?'] }.to_h
-          meta_data[:filters] = JSON.dump(filters)
+          filter = command["filter"].map { |key, _v| [key, '?'] }.to_h
+          meta_data[:filter] = JSON.dump(filter)
         end
       end
       meta_data[:message] = event.message if defined?(event.message)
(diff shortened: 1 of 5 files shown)


lib/bugsnag/integrations/mongo.rb
```diff
@@ -48,13 +49,31 @@ def leave_mongo_breadcrumb(event_name, event)
         :command_name => event.command_name,
         :database_name => event.database_name,
         :operation_id => event.operation_id,
-        :request_id => event.request_id
+        :request_id => event.request_id,
+        :duration => event.duration
       }
-      meta_data[:duration] = event.duration if defined?(event.duration)
+      if command = pop_command(event.request_id)
+        meta_data[:collection] = command[event.command_name]
```
tobyhs: For `getMore` commands, the collection name is in `command['collection']` (and not `command['getMore']`)

Cawllec: From looking through their test specifications [here](https://github.com/mongodb/mongo-ruby-driver/tree/master/spec/spec_tests/data/command_monitoring), it looks like `getMore` is the only one.

lib/bugsnag/integrations/mongo.rb
```diff
@@ -48,13 +49,47 @@ def leave_mongo_breadcrumb(event_name, event)
...
+    def leave_command(event)
+      event_commands[event.request_id] = event.command
+    end
+
+    ##
+    # Removes and retrieves a stored command from the request data
+    #
+    # @param request_id [String] the id of the mongo_ruby_driver event
+    #
+    # @return [Hash|nil] the requested command, or nil if not found
```
tobyhs: `[Hash, nil]` (as I don't think a `|` is valid in a yardoc type description)

PR #521, opened 2019-01-04
Breadcrumbs: Additional Rails ActiveRecord data

Cawllec:
## Goal
- Adds the capture of the `event_id` for each event
- Captures the SQL string of ActiveRecord queries <- Now split out into separate (later) task.
- Captures binding keys of ActiveRecord queries, with values explicitly redacted

### DIFF
--- a/features/fixtures/rails3/app/app/controllers/breadcrumbs_controller.rb
+++ b/features/fixtures/rails3/app/app/controllers/breadcrumbs_controller.rb
@@ -1,4 +1,3 @@
-require 'pp'
 class BreadcrumbsController < ApplicationController
   def handled
     Bugsnag.notify("Request breadcrumb")
--- a/features/fixtures/rails3/app/app/controllers/breadcrumbs_controller.rb
+++ b/features/fixtures/rails3/app/app/controllers/breadcrumbs_controller.rb
@@ -1,3 +1,4 @@
+require 'pp'
 class BreadcrumbsController < ApplicationController
   def handled
     Bugsnag.notify("Request breadcrumb")
@@ -5,7 +6,7 @@ class BreadcrumbsController < ApplicationController
   end
 
   def sql_breadcrumb
-    User.take
+    User.where(:email => "foo").as_json
     Bugsnag.notify("SQL breadcrumb")
     render json: {}
   end
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -82,7 +82,6 @@ module Bugsnag
           binds = data[:binds].map { |bind| defined?(bind.name) ? [bind.name, '?'] : nil }
           binds.select! { |bind| bind }
           filtered_data[:bindings] = JSON.dump(binds.to_h) unless binds.empty?
-          filtered_data[:sql] = data[:sql].gsub(/(=|>|<|=>|=<|IN) (\S+)/, '\1 "?"')
         end
         Bugsnag.leave_breadcrumb(
           event[:message],
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -1,5 +1,6 @@
 # Rails 3.x hooks
 
+require "json"
 require "rails"
 require "bugsnag"
 require "bugsnag/middleware/rails3_request"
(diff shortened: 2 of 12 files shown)


lib/bugsnag/integrations/railtie.rb
```diff
@@ -73,9 +74,15 @@ class Railtie < ::Rails::Railtie
     # @api private
     # @param event [Hash] details of the event to subscribe to
     def event_subscription(event)
-      ActiveSupport::Notifications.subscribe(event[:id]) do |*, data|
+      ActiveSupport::Notifications.subscribe(event[:id]) do |*, event_id, data|
         filtered_data = data.slice(*event[:allowed_data])
         filtered_data[:event_name] = event[:id]
+        filtered_data[:event_id] = event_id
+        if event[:id] == "sql.active_record"
```
tobyhs: Can you add test(s) for this event type? (especially since you're adding more logic around this specific case)

lib/bugsnag/integrations/railtie.rb
```diff
@@ -73,9 +74,15 @@ class Railtie < ::Rails::Railtie
...
+          binds = data[:binds].map { |bind| defined?(bind.name) ? [bind.name, '?'] : nil }
```
tobyhs: I don't think `binds` is used when using the mysql2 activerecord adapter, so queries (in the `:sql` entry of the data/payload) won't be redacted for Rails apps using mysql2.

For reference, I was testing a Rails 5.1 app, and something like `Post.where(title: 'foobar')` resulted in the following notification payload:
```
{:sql=>"SELECT  `posts`.* FROM `posts` WHERE `posts`.`title` = 'foobar' LIMIT 11", :name=>"Post Load", :binds=>[], :type_casted_binds=>[], :statement_name=>nil, :connection_id=>70150552673420}
```

Cawllec: So long as the query quotes stay the same as in that sql string (which I'll verify), it should be something fairly easily replaceable.

features/fixtures/rails3/app/app/controllers/breadcrumbs_controller.rb
```diff
@@ -1,11 +1,12 @@
+require 'pp'
```
tobyhs: Remove this line

lib/bugsnag/integrations/railtie.rb
```diff
@@ -73,9 +74,16 @@ class Railtie < ::Rails::Railtie
...
+          binds.select! { |bind| bind }
+          filtered_data[:bindings] = JSON.dump(binds.to_h) unless binds.empty?
+          filtered_data[:sql] = data[:sql].gsub(/(=|>|<|=>|=<|IN) (\S+)/, '\1 "?"')
```
tobyhs: This gsub isn't working as I would expect for `IN` comparison (the 2nd element is still left in):
```
irb(main):001:0> puts "SELECT  `posts`.* FROM `posts` WHERE `posts`.`title` IN ('abc', 'def')".gsub(/(=|>|<|=>|=<|IN) (\S+)/, '\1 "?"')
SELECT  `posts`.* FROM `posts` WHERE `posts`.`title` IN "?" 'def')
```

Also, I'm sure there are other cases that aren't covered. See https://github.com/percona/percona-toolkit/blob/4c28e71092f1405cc75c5c4a1e28cfb9be94c570/bin/pt-fingerprint#L1635 to get an idea of how Percona Toolkit's pt-fingerprint approaches it (it isn't super simple).

lib/bugsnag/integrations/railtie.rb
```diff
@@ -73,9 +74,15 @@ class Railtie < ::Rails::Railtie
...
     def event_subscription(event)
-      ActiveSupport::Notifications.subscribe(event[:id]) do |*, data|
+      ActiveSupport::Notifications.subscribe(event[:id]) do |*, event_id, data|
         filtered_data = data.slice(*event[:allowed_data])
         filtered_data[:event_name] = event[:id]
+        filtered_data[:event_id] = event_id
+        if event[:id] == "sql.active_record"
+          binds = data[:binds].map { |bind| defined?(bind.name) ? [bind.name, '?'] : nil }
+          binds.select! { |bind| bind }
+          filtered_data[:bindings] = JSON.dump(binds.to_h) unless binds.empty?
```
tobyhs: I think we should call this `:binds` instead of `:bindings` to be consistent with ActiveRecord's naming.

PR #522, opened 2019-01-08
Breadcrumbs/Mongo integration: Handle 'any_of' cases

Cawllec:
## Goal
Enables mongo breadcrumb filters to include nested keys, such as in `any_of` cases.

### DIFF
--- a/lib/bugsnag/integrations/mongo.rb
+++ b/lib/bugsnag/integrations/mongo.rb
@@ -74,7 +74,7 @@ module Bugsnag
     #
     # @return [Hash] the filtered hash
     def sanitize_filter_hash(filter_hash, depth = 0)
-      filter_hash.each_with_object(Hash.new) do |args, output|
+      filter_hash.each_with_object({}) do |args, output|
         key, value = *args
         output[key] = sanitize_filter_value(value, depth)
       end
--- a/lib/bugsnag/integrations/mongo.rb
+++ b/lib/bugsnag/integrations/mongo.rb
@@ -74,9 +74,10 @@ module Bugsnag
     #
     # @return [Hash] the filtered hash
     def sanitize_filter_hash(filter_hash, depth = 0)
-      filter_hash.map do |key, value|
-        [key, sanitize_filter_value(value, depth)]
-      end.to_h
+      filter_hash.each_with_object(Hash.new) do |args, output|
+        key, value = *args
+        output[key] = sanitize_filter_value(value, depth)
+      end
     end
 
     ##
--- a/lib/bugsnag/integrations/mongo.rb
+++ b/lib/bugsnag/integrations/mongo.rb
@@ -10,6 +10,7 @@ module Bugsnag
     MONGO_MESSAGE_PREFIX = "Mongo query "
     MONGO_EVENT_PREFIX = "mongo."
     MONGO_COMMAND_KEY = :bugsnag_mongo_commands
+    MAX_FILTER_DEPTH = 5
 
     ##
     # Listens to the 'started' event, storing the command for later usage
@@ -56,7 +57,7 @@ module Bugsnag
         collection_key = event.command_name == "getMore" ? "collection" : event.command_name
         meta_data[:collection] = command[collection_key]
         unless command["filter"].nil?
-          filter = sanitize_filter(command["filter"])
+          filter = sanitize_filter_hash(command["filter"])
           meta_data[:filter] = JSON.dump(filter)
         end
       end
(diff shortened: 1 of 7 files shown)


lib/bugsnag/integrations/mongo.rb
```diff
@@ -65,6 +65,23 @@ def leave_mongo_breadcrumb(event_name, event)
...
+    ##
+    # Removes values from filter hashes, replacing them with '?'
+    #
+    # @param filter_hash [Hash] the filter hash for the mongo transaction
+    #
+    # @return [Hash] the filtered hash
+    def sanitize_filter(filter_hash)
+      filter_hash.map do |key, value|
+        if value && value.is_a?(Array)
+          filtered_or = value.map { |filter| sanitize_filter(filter) }
```
tobyhs: This variable could be slightly misleading because an Array value doesn't necessarily mean it was an `$or` (it could have been `$and` or something else).

lib/bugsnag/integrations/mongo.rb
```diff
@@ -65,6 +65,23 @@ def leave_mongo_breadcrumb(event_name, event)
       Bugsnag.leave_breadcrumb(message, meta_data, Bugsnag::Breadcrumbs::PROCESS_BREADCRUMB_TYPE, :auto)
     end
 
+    ##
+    # Removes values from filter hashes, replacing them with '?'
+    #
+    # @param filter_hash [Hash] the filter hash for the mongo transaction
+    #
+    # @return [Hash] the filtered hash
+    def sanitize_filter(filter_hash)
```
tobyhs: This method fails on queries such as `SomeModel.where(tags: ['a', 'b'])`:
```
irb(main):001:0> def sanitize_filter(filter_hash)
irb(main):002:1>   filter_hash.map do |key, value|
irb(main):003:2*     if value && value.is_a?(Array)
irb(main):004:3>       filtered_or = value.map { |filter| sanitize_filter(filter) }
irb(main):005:3>       [key, filtered_or]
irb(main):006:3>     else
irb(main):007:3*       [key, '?']
irb(main):008:3>     end
irb(main):009:2>   end.to_h
irb(main):010:1> end
=> :sanitize_filter
irb(main):011:0> sanitize_filter({'tags' => [1, 2]})
NoMethodError: undefined method `map' for 1:Integer
Did you mean?  tap
	from (irb):2:in `sanitize_filter'
	from (irb):4:in `block (2 levels) in sanitize_filter'
	from (irb):4:in `map'
	from (irb):4:in `block in sanitize_filter'
	from (irb):2:in `each'
	from (irb):2:in `map'
	from (irb):2:in `sanitize_filter'
	from (irb):11
```

I would change the method/algorithm to something like:
```ruby
def sanitize_filter_element(element)
  if element.is_a?(Array)
    element.map { |e| sanitize_filter_element(e) }
  elsif element.is_a?(Hash)
    element.map { |k, v| [k, sanitize_filter_element(v)] }.to_h
  else
    '?'
  end
end
```

The only concern I have about the above is that there could be giant arrays for `$in`, but I'll pull the old college textbook approach of writing "This is left as an exercise for the reader".

spec/integrations/mongo_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+    describe "#started" do
+      it "calls #leave_command with the event" do
+        event = double
+        expect(subscriber).to receive(:leave_command).with(event)
+        subscriber.started(event)
+      end
+    end
+
+    describe "#succeeded" do
+      it "calls #leave_mongo_beradcrumb with the event_name and event" do
```
tobyhs: leave_mongo_breadcrumb (spelling)

spec/integrations/mongo_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+              :duration => "123.456",
+              :collection => "collection_name_command"
+            },
+            "process",
+            :auto
+          )
+          subscriber.send(:leave_mongo_breadcrumb, event_name, event)
+        end
+
+        it "adds the correct colleciton name for 'getMore' commands" do
```
tobyhs: collection (spelling)

spec/integrations/mongo_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+            "Mongo query #{event_name}",
+            {
+              :event_name => "mongo.#{event_name}",
+              :command_name => "command",
+              :database_name => "database",
+              :operation_id => "1234567890",
+              :request_id => "123456",
+              :duration => "123.456",
+              :collection => "collection_name_command",
+              :filter => "{\"a\":\"?\",\"b\":\"?\",\"$or\":[{\"c\":\"?\"},{\"d\":\"?\"}]}"
```
tobyhs: I would use single quotes for the outer quotes here so that the blackslashes aren't needed and it is easier to read this string literal

spec/integrations/mongo_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+            :auto
+          )
+          subscriber.send(:leave_mongo_breadcrumb, event_name, event)
+        end
+      end
+    end
+
+    describe "#sanitize_filter_hash" do
+      it "calls into #sanitize_filter_value with the value from each {k,v} pair" do
+        expect(subscriber).to receive(:sanitize_filter_value).with(1, 0).ordered.and_return('?')
```
tobyhs: I would omit all your `sanitize_filter_value` method/message expectations since your equality expectations should be good enough. Method/message expectations tend to be for things that have side effects that you don't want run (e.g. remote/network calls).

spec/integrations/mongo_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+      end
+
+      it "returns the command given a request_id" do
+        expect(subscriber.send(:pop_command, request_id)).to eq(command)
+      end
+
+      it "removes the command from the request_data" do
+        subscriber.send(:pop_command, request_id)
+        command_hash = Bugsnag.configuration.request_data[Bugsnag::MongoBreadcrumbSubscriber::MONGO_COMMAND_KEY]
+        expect(command_hash).not_to include(request_id => command)
```
tobyhs: I would change this to `not_to have_key(request_id)` as the key shouldn't be in at all and the value is not relevant to the test.

spec/integrations/mongo_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+        alias_method :old_require, :require
+        def require(path)
+          old_require(path) unless path == 'mongo'
+        end
+      end
+    end
+  end
+
+  it "should subscribe to the mongo monitoring service" do
+    expect(::Mongo::Monitoring::Global).to receive(:subscribe) do |*args|
```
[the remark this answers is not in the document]
Cawllec: At the point this is run, the `::Bugsnag::MongoBreadcrumbSubscriber` doesn't exist, so I've split the difference by leaving it as a block and checking `command` and `subscriber` separately, but making an `instance_of` call rather than checking `respond_to?`s.

PR #523, opened 2019-01-11
Capture sessions automatically by default

Cawllec:
## Goal

Enable automatic session capture by default, and more strictly enforce endpoint setting.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -305,6 +305,7 @@ Metrics/ModuleLength:
   Max: 125
   Exclude:
     - 'lib/bugsnag/helpers.rb'
+    - 'lib/bugsnag.rb'
 
 # Offense count: 11
 Metrics/PerceivedComplexity:
--- a/features/fixtures/rails3/app/app/controllers/session_tracking_controller.rb
+++ b/features/fixtures/rails3/app/app/controllers/session_tracking_controller.rb
@@ -6,15 +6,19 @@ class SessionTrackingController < ActionController::Base
   end
 
   def initializer
-    Bugsnag.session_tracker.send_sessions()
+    Bugsnag.session_tracker.send_sessions
     render json: {}
   end
 
-  def after
-    Bugsnag.configure do |conf|
-      conf.auto_capture_sessions = true
-    end
-    Bugsnag.session_tracker.send_sessions()
+  def manual
+    Bugsnag.start_session
+    Bugsnag.session_tracker.send_sessions
+    render json: {}
+  end
+
+  def hundred
+    (0...100).each { Bugsnag.start_session }
+    Bugsnag.session_tracker.send_sessions
     render json: {}
   end
 end
--- a/features/fixtures/rails3/app/app/controllers/session_tracking_controller.rb
+++ b/features/fixtures/rails3/app/app/controllers/session_tracking_controller.rb
@@ -16,7 +16,7 @@ class SessionTrackingController < ActionController::Base
     render json: {}
   end
 
-  def hundred
+  def multi_sessions
     (0...100).each { Bugsnag.start_session }
     Bugsnag.session_tracker.send_sessions
     render json: {}
--- a/features/rails_features/auto_capture_sessions.feature
+++ b/features/rails_features/auto_capture_sessions.feature
@@ -5,45 +5,18 @@ Background:
   And I set environment variable "APP_PATH" to "/usr/src"
   And I configure the bugsnag endpoint
 
-Scenario Outline: Auto_capture_sessions can be set to true in the initializer
+Scenario Outline: 100 session calls results in 100 sessions
   Given I set environment variable "RUBY_VERSION" to "<ruby_version>"
-  And I set environment variable "BUGSNAG_AUTO_CAPTURE_SESSIONS" to "true"
+  And I set environment variable "BUGSNAG_AUTO_CAPTURE_SESSIONS" to "false"
   And I start the service "rails<rails_version>"
   And I wait for the app to respond on port "6128<rails_version>"
-  When I navigate to the route "/session_tracking/initializer" on port "6128<rails_version>"
+  When I navigate to the route "/session_tracking/hundred" on port "6128<rails_version>"
   Then I should receive a request
   And the request is a valid for the session tracking API
   And the request used the "Ruby Bugsnag Notifier" notifier
   And the "Bugsnag-API-Key" header equals "a35a2a72bd230ac0aa0f52715bbdc6aa"
   And the sessionCount "startedAt" is a timestamp
-
-  Examples:
-    | ruby_version | rails_version |
-    | 2.0          | 3             |
-    | 2.1          | 3             |
-    | 2.2          | 3             |
-    | 2.2          | 4             |
-    | 2.2          | 5             |
-    | 2.3          | 3             |
-    | 2.3          | 4             |
-    | 2.3          | 5             |
-    | 2.4          | 3             |
-    | 2.4          | 5             |
-    | 2.5          | 3             |
-    | 2.5          | 5             |
-
-Scenario Outline: Auto_capture_sessions can be set to true after the initializer
-  Given I set environment variable "RUBY_VERSION" to "<ruby_version>"
-  And I start the service "rails<rails_version>"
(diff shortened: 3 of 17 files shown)


.rubocop_todo.yml
```diff
@@ -305,6 +305,7 @@ Metrics/ModuleLength:
   Max: 125
   Exclude:
     - 'lib/bugsnag/helpers.rb'
+    - 'lib/bugsnag.rb'
```
Cawllec: Note - this exclusion was originally approved in the `breadcrumbs/base` branch.

features/fixtures/rails3/app/app/controllers/session_tracking_controller.rb
```diff
@@ -6,15 +6,19 @@ def index
...
-      conf.auto_capture_sessions = true
-    end
-    Bugsnag.session_tracker.send_sessions()
+  def manual
+    Bugsnag.start_session
+    Bugsnag.session_tracker.send_sessions
+    render json: {}
+  end
+
+  def hundred
```
tobyhs: I would name this `many` or `multiple` or something like that so the name still makes sense if you change the number of sessions started.

lib/bugsnag/configuration.rb
```diff
@@ -35,12 +34,26 @@ class Configuration
...
+    attr_reader :notify_endpoint
+    alias :endpoint :notify_endpoint
+
+    ##
+    # @return [String] URL session notifications will be delivered to
+    attr_reader :session_endpoint
+
+    ##
+    # @return [Boolean] whether any sessions types will be delivered
+    attr_reader :send_sessions
```
tobyhs: Having a boolean setting/option/attribute with the same name as a public method in `SessionTracker` (`send_sessions`) might be slightly confusing. It might be preferable to rename this option to something like `sessions_enabled` or `session_tracking_enabled` instead.

spec/bugsnag_spec.rb
```diff
@@ -17,6 +17,47 @@
...
+    let(:custom_notify_endpoint) { "Custom notify endpoint" }
+    let(:custom_session_endpoint) { "Custom session endpoint" }
+
+    it "does nothing for default endpoints or if both endpoints are set" do
+      expect(Bugsnag.configuration).not_to receive(:warn)
+      Bugsnag.send(:check_endpoint_setup)
+
+      Bugsnag.configuration.set_endpoints(custom_notify_endpoint, custom_session_endpoint)
+      Bugsnag.send(:check_endpoint_setup)
+    end
```
tobyhs: Add blank lines between `it` blocks

spec/session_tracker_spec.rb
```diff
@@ -60,11 +72,21 @@
...
   end
 
+  it 'will not create sessions if Configuration.send_sessions is false' do
+    Bugsnag.configure do |conf|
+      conf.set_endpoints("http://localhost:#{server.config[:Port]}", nil)
+    end
+    expect(Bugsnag.configuration.send_sessions).to eq(false)
+    expect(Bugsnag.session_tracker.session_counts.size).to eq(0)
+    Bugsnag.start_session
+    # expect(Bugsnag.session_tracker.session_counts.size).to eq(0)
```
tobyhs: Did you intend to uncomment this out?

spec/session_tracker_spec.rb
```diff
@@ -2,6 +2,14 @@
 require 'webrick'
 require 'spec_helper'
 require 'json'
+require 'concurrent'
+
+# Enable reset of session_counts between each example
+module Bugsnag
+  class SessionTracker
+    attr_accessor :session_counts
```
tobyhs: I think you can avoid opening/monkey-patching this class by having a `before` block call `Bugsnag.session_tracker.send_sessions` to flush `session_counts`, although you will probably need to set the delivery method to synchronous and clear `queue` right after the `send_sessions` call.

features/rails_features/auto_capture_sessions.feature
```diff
@@ -5,45 +5,18 @@ Background:
...
-  When I navigate to the route "/session_tracking/after" on port "6128<rails_version>"
-  Then I should receive 0 requests
-  When I navigate to the route "/session_tracking/initializer" on port "6128<rails_version>"
+  When I navigate to the route "/session_tracking/hundred" on port "6128<rails_version>"
   Then I should receive a request
   And the request is a valid for the session tracking API
   And the request used the "Ruby Bugsnag Notifier" notifier
   And the "Bugsnag-API-Key" header equals "a35a2a72bd230ac0aa0f52715bbdc6aa"
   And the sessionCount "startedAt" is a timestamp
+  And the sessionCount "sessionsStarted" equals 100
```
tobyhs: Is this a little prone to non-deterministic failures because the 100 sessions could span across 2 different minutes?

features/rails_features/auto_capture_sessions.feature
```diff
@@ -5,45 +5,18 @@ Background:
   And I set environment variable "APP_PATH" to "/usr/src"
   And I configure the bugsnag endpoint
 
-Scenario Outline: Auto_capture_sessions can be set to true in the initializer
```
tobyhs: Should this test be left in (and perhaps adjusted a little) because it tests auto-capturing sessions (which your new tests don't because the new controller actions call `Bugsnag.start_session` manually)?

spec/session_tracker_spec.rb
Cawllec: To be honest, as long as the monkey-patching won't interfere with the tests I don't have a problem with it.

features/steps/ruby_notifier_steps.rb
```diff
@@ -22,4 +22,10 @@
   steps %Q{
     the "#{element}" of stack frame #{frame_index} equals #{value}
   }
+end
+
+Then(/^the total sessionStarted count equals (\d+)(?: for request (\d+))?$/) do |value, request_index|
+  session_counts = read_key_path(find_request(request_index)[:body], "sessionCounts")
+  total_count = session_counts.inject(0) { |count, session| count += session["sessionsStarted"] }
+  assert_equal(total_count, value)
```
tobyhs: Flip these arguments (`(value, total_count)`); the first parameter is `expected` and the second is `actual`.

spec/session_tracker_spec.rb
tobyhs: A risk of monkey patching like this is that someone might introduce app code that uses the extra behavior (`Bugsnag::SessionTracker#session_counts=` in this case) and rely on the unit tests to make sure things work, which will be not as accurate as the tests introduce an extra method while the app code doesn't have this extra method.

Alternatively, you can do something similar to the global `before(:each)` block in spec_helper.rb and have `Bugsnag.instance_variable_set(:@session_tracker, Bugsnag::SessionTracker.new)` in a `before` block in this file (which has the benefit that anything in `Bugsnag.session_tracker` is reset if more instance variables are added to `Bugsnag::SessionTracker` later).

spec/configuration_spec.rb
```diff
@@ -22,10 +22,93 @@
...
+    it "set notify_endpoint and session_endpoint" do
+      subject.set_endpoints(custom_notify_endpoint, custom_session_endpoint)
+      expect(subject.notify_endpoint).to eq(custom_notify_endpoint)
+      expect(subject.session_endpoint).to eq(custom_session_endpoint)
+    end
+  end
 
-    it "should have sensible defaults for session tracking" do
-      expect(subject.session_endpoint).to eq("https://sessions.bugsnag.com")
+  describe "disable_sessions" do
```
tobyhs: `#disable_sessions` (missing `#`)

spec/session_tracker_spec.rb
Cawllec: That makes sense, and I quite like that last suggestion, more so than relying on `send_sessions`.

tobyhs: Shouldn't you remove this monkey patch and the `require 'concurrent'` now that you're using `Bugsnag.instance_variable_set(:@session_tracker, Bugsnag::SessionTracker.new)`?

Also, I would put it in a `before(:each)` instead, then have an `after(:all)` so you have something fresh before the first test in this file also.

PR #526, opened 2019-01-17
Fix flaky session-tracking test

Cawllec:
## Goal
Stabilise test introduced with the auto-session-tracking feature


spec/session_tracker_spec.rb
```diff
@@ -4,13 +4,6 @@
 require 'json'
 require 'concurrent'
```
tobyhs: I think you can also remove this line

PR #530, opened 2019-01-22
Fix unnecessary meta_data being logged during breadcrumb validation

### LINKED ISSUE #528: Latest version is logging user data unnecessarily
### Description
We just upgraded to version 6.11 and we noticed that our logs were filling up with warnings:

```
2019-01-21T12:42:54.953605+00:00 app[web.1]: W, [2019-01-21T12:42:54.953494 #14]  WARN -- [Bugsnag]: [bd099a85-85f2-4fb8-886c-3c307e773b6a] Breadcrumb Read cache meta_data key:[:views, "runners/_streaks:39bb5fecc026ce9a483453b23996190c", #<Runner id: 1234, name: "Zara", slug: "zara-261b5321-12c5-4b0a-bbae-12345678", updated_at: "2017-05-21 15:20:49", home_area_id: 50, had_training_call_at: nil, user_id: 43135, avatar_data: nil, tag_list: nil>] has been dropped for having an invalid data type
```

Looks like they are being caused by the new breadcrumb feature?
https://github.com/bugsnag/bugsnag-ruby/blob/0466dd2a654cc1adbd1e1bb2fb7ba128fae3ba85/lib/bugsnag/breadcrumbs/validator.rb#L29

Should this not be a `debug` rather than a `warn`?

Cawllec:
## Goal

Alters breadcrumb validation logs:
- Reduces the log-level to `debug`
- No longer logs the value of meta_data when rejected, only the key

In addition:
- Removes `key` parameter from cache breadcrumbs due to potentially being an un-breadcrumbable type

Fixes #528


lib/bugsnag/breadcrumbs/validator.rb
```diff
@@ -26,21 +26,21 @@ def validate(breadcrumb)
         if valid_meta_data_type?(v)
           true
         else
-          @configuration.warn("Breadcrumb #{breadcrumb.name} meta_data #{k}:#{v} has been dropped for having an invalid data type")
+          @configuration.debug("Breadcrumb #{breadcrumb.name} meta_data #{k} has been dropped for having an invalid data type")
```
tobyhs: Should you include `v.class` in the log message? (as that shouldn't be sensitive)

PR #531, opened 2019-02-08
If a custom object responds to `id` method, show the id and class value, instead of showing "[OBJECT]" in error reports

manojmj92:
## Goal

When bugsnag prints out the arguments passed to a job, if the arguments are custom objects (like a Struct, or an ActiveRecord object), Bugsnag only prints out "[OBJECT]" in its place, instead of providing any details, like

```json
{
  "class": "SubcriptionRenewalJob",
  "args": {
    "user": "[OBJECT]",
    "account": "[OBJECT]",
    "setting": "[OBJECT]",
    "subscription": "[OBJECT]"
  }
}
```

While bugsnag does this for the right reason (we should not show details of the whole object using `inspect`, because of PII), for developers who have not written a `to_s` in the model, just seeing `"[OBJECT]"` in place of the actual object details becomes a nightmare. Such errors become practically impossible to debug because, well, you do not know which object caused the issue.

This PR is aimed at solving this problem to some extent, ie, at least in case of ActiveRecord objects.

## Design

All ActiveRecord objects have an `id` method on it by default. If the custom object does not respond to `to_s` and we cannot show it on bugsnag due to data-sensitivity issue, we can still show the `id` of the object.

Reasons:

1. ID is technically not a sensitive data point, like say `email` or `address` (which could be exposed if we give out the whole value of `inspect` on an object)
2. For developers debugging the issue, just knowing the ID of the object should suffice to go back and fetch the record from the database, for debugging purposes.

## Result

Bugsnag will now show arguments as

```json
{
  "class": "SubcriptionRenewalJob",
  "args": {
    "user": "[OBJECT]: [Class]: User [ID]: 23",
    "account": "[OBJECT]: [Class]: Account [ID]: 18",
    "setting": "[OBJECT]: [Class]: Setting [ID]: 7",
    "subscription": "[OBJECT]: [Class]: Subscription [ID]: 88",
  }
}
```
### Added

Added condition to check if the object responds to `id` method. If so, the `id` of the object is being returned as the value of the argument.

## Tests

Added tests to assert that for custom objects that respond to the `id` method, the `id` of the object is returned as the value of the object, rather than the string `"[OBJECT]"`

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/cleaner.rb
+++ b/lib/bugsnag/cleaner.rb
@@ -7,6 +7,7 @@ module Bugsnag
     RECURSION = '[RECURSION]'.freeze
     OBJECT = '[OBJECT]'.freeze
     RAISED = '[RAISED]'.freeze
+    OBJECT_WITH_ID_AND_CLASS = '[OBJECT]: [Class]: %{class_name} [ID]: %{id}'.freeze
 
     def initialize(filters)
       @filters = Array(filters)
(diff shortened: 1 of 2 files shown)


imjoehaines:
Thanks for the PR @manojmj92 — I've fixed the conflict and will merge this in https://github.com/bugsnag/bugsnag-ruby/pull/605 🙂

PR #535, opened 2019-03-30
Fix v6 regression in Resque integration by setting context explicitly

oehlschl:
## Goal

- fix a regression introduced in v6.0 (Notification -> Report refactor) where the Resque job class and queue were no longer being set and shown in the Bugsnag dashboard
   - in v5, Resque context was sent to `auto_notify` as one of the `@overrides`, and was eventually set on the Notification during `deliver`: https://github.com/bugsnag/bugsnag-ruby/blob/v5.5.0/lib/bugsnag/notification.rb#L242-L242
   - in v6, however, Resque context is set in the `meta_data`, and meta_data keys do not override public Report setters in the same way
- the goal of this PR is fixing what seems like a regression in the Bugsnag Resque integration and showing a more helpful context in the Bugsnag dashboard than `resque:work` 

## Design

- this change is intended to be minimally invasive and keeps the context as part of the meta_data, while also setting it explicitly

## Changeset

- changed only `Bugsnag::Resque#save`

## Tests

- updated specs
- test locally using our own application's Bugsnag integration
   - asserted that after the change, the Bugsnag dashboard showed the Resque task name and queue as context rather than `resque:work`

Before (example):
Screen Shot 2019-03-29 at 21 03 09

After (admittedly a different example, but it still shows the change):
Screen Shot 2019-03-29 at 21 02 44

## Discussion

- n/a, but please LMK if this is too breaking, in that it does change the v6 context for Resque (admittedly fixing it back to what it was in v5)

### Linked issues

- n/a, but please LMK if you want me to file an issue for the regression

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/integrations/resque.rb
+++ b/lib/bugsnag/integrations/resque.rb
@@ -43,7 +43,8 @@ module Bugsnag
           :type => Bugsnag::Report::UNHANDLED_EXCEPTION_MIDDLEWARE,
           :attributes => FRAMEWORK_ATTRIBUTES
         }
-        report.meta_data.merge!({:context => "#{payload['class']}@#{queue}", :payload => payload})
+        report.context = "#{payload['class']}@#{queue}"
+        report.meta_data[:payload] = payload
       end
     end
   end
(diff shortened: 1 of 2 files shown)


oehlschl:
Adding some people I've seen in other PRs, for visibility 

@snmaynard @Cawllec

PR #538, opened 2019-04-26
Fix e2e fixtures and 1.9.3 unit tests

Cawllec:
## Goal

- Adds gem requirements for Ruby 1.9.3 to ensure unit tests run
- Fixes SQLite3 version in Rails fixtures for e2e tests


features/fixtures/rails5/app/Gemfile
```diff
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.0.0'
```
tobyhs: I think Rails 5.2.3 has the fix for the sqlite3 version dependency, but this dependency here won't go up to 5.2. We should probably consider adding or changing this some day.

PR #544, opened 2019-06-10
Allow vendor_path to be configured

jnraine:
## Goal

By default, lines within `vendor/` or `.bundle/` are consider “out of project”. This is a good default for most Rails projects but there are other directories we’d also like to ignore. For example, patches to gems or the framework. If left as “in project”, these lines cause multiple issues to be grouped together because the first line in the stacktrace always points to the same patch line.

Right now, we’ve side stepped this by reaching into the gem and redefining `Bugsnag::Stacktrace::VENDOR_PATH` but it would be wonderful if this was a configurable option like any other.

### Added

`vendor_path` can now be configured.

This allows developers to set the vendor path regular expression, used to determine whether or not a line in the stacktrace appears in the project stacktrace or the full trace.

```ruby
Bugsnag.configure do |config|
  config.vendor_path = /(^(vendor\/|\.bundle\/|extensions\/))/
end
```

## Tests

- Test for the configuration setter and getter were added.
- Tests to ensure the new configuration option was respected was added for `Bugsnag::Stacktrace`.

### Outstanding Questions

Should the name of the configuration option be changed? Instead of `vendor_path`, perhaps `project_file_pattern` or similar.

## Review

- [ ] ~~Commented on code changes inline explain the reasoning behind the approach~~ n/a
  - [ ] Final review


jnraine:
CI failure seems unrelated to this changeset but I can't explain why master is passing while this branch isn't:

```
$ bundle _${BUNDLE_VERSION}_ install --with "$GEMSETS" --binstubs
Fetching https://github.com/bugsnag/maze-runner
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies....
Installing rake 12.3.2
Installing concurrent-ruby 1.1.5
Installing minitest 5.11.3
Installing thread_safe 0.3.6
Installing addressable 2.3.8
Installing coderay 1.1.2
Installing connection_pool 2.2.2
Installing safe_yaml 1.0.5
Installing diff-lcs 1.3
Installing hashdiff 0.4.0
Installing method_source 0.9.2
Installing rack 2.0.7
Installing rdoc 5.1.0
Installing redis 4.1.2                                    -----
Gem::InstallError: redis requires Ruby version >= 2.3.0. <----- redis 4.1.2 doesn't work with Ruby 2.2
Installing rspec-support 3.8.0                            -----
Using bundler 1.12.0
Installing i18n 1.4.0
Using bugsnag 6.11.1 from source at `.`
Installing tzinfo 1.2.5
Installing crack 0.4.3
Installing pry 0.12.2
Installing rack-protection 2.0.5
An error occurred while installing redis (4.1.2), and Bundler cannot continue.
Make sure that `gem install redis -v '4.1.2'` succeeds before bundling.
The command "bundle _${BUNDLE_VERSION}_ install --with "$GEMSETS" --binstubs" failed and exited with 5 during .
```
🤔

tobyhs:
I fixed the Ruby 2.2 + redis gem issue in https://github.com/bugsnag/bugsnag-ruby/pull/546 so you can merge master into your branch to fix it if you want. However, new features go into the `next` branch, so you might want to wait until that branch is brought up to date and then we'll try switching the base branch of this pull request to `next`.

lib/bugsnag/configuration.rb
```diff
@@ -173,6 +177,13 @@ def should_notify_release_stage?
       @release_stage.nil? || @notify_release_stages.nil? || @notify_release_stages.include?(@release_stage)
     end
 
+    ##
+    # Regex used to mark stacktrace lines as within the project. Lines marked
+    # as "out of project" will only appear in the full trace.
+    def vendor_path
```
tobyhs: Add an `@return` comment (or not if you choose to use `attr_accessor` instead and set it to the default in initialize)

lib/bugsnag/configuration.rb
```diff
@@ -173,6 +177,13 @@ def should_notify_release_stage?
...
+      @vendor_path || Bugsnag::Stacktrace::VENDOR_PATH
```
tobyhs: Consider moving/renaming `Bugsnag::Stacktrace::VENDOR_PATH` to `Bugsnag::Configuration::DEFAULT_VENDOR_PATH`, as it is now a default, and it might be easier for users to find/discover this constant in `Bugsnag::Configuration` in case they need to use this regular expression to add to it for a new vendor path regular expression (maybe using something like `Regexp.union`).

spec/configuration_spec.rb
```diff
@@ -396,4 +396,15 @@ def debug(name, &block)
       expect(second_array).to eq([1, 2])
     end
   end
+
+  describe "vendor_path" do
```
tobyhs: `#vendor_path` (with the `#` since it is an instance method)

spec/stacktrace_spec.rb
```diff
@@ -87,4 +87,43 @@
       })
     }
   end
+
+  describe "#vendor_cache" do
```
tobyhs: I don't understand this doc string, as there isn't a `vendor_cache` instance method, and the tests don't appear to be related to caching.

spec/configuration_spec.rb
```diff
@@ -396,4 +396,15 @@ def debug(name, &block)
...
+    it "should have the default vendor path" do
```
tobyhs: I prefer to avoid "should" in doc strings (https://github.com/rubocop-hq/rspec-style-guide#should-in-example-docstrings), mainly to avoid repeating it a lot (similar comment for line 405)

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Enhancements
+
+* Add option to configure what file paths are included in the project stacktrace (`vendor_path`)
```
tobyhs: This doesn't seem clear, as all the file paths are usually included in the stacktrace. I would reword this to something like "Add option (`vendor_path`) to adjust filename matching for marking stackframes as in-project".

tobyhs:
I updated the next branch and changed the base branch of this pull request to next. @jnraine Can you update this when you get the chance?

rafaelfranca:
This option was present in 5.5 and was removed without giving any reason. https://github.com/bugsnag/bugsnag-ruby/pull/253. It used to allow arrays, and I believe it still should since you can have more than 1 vendor path.

jnraine:
@tedyangx/@mctaylorpants, can you take this to completion? It'll let you drop a monkey patch from the codebase. 😄

mctaylorpants:
@jnraine yep we can take a look! Are you able to give us write access to your fork?

jnraine:
@mctaylorpants Done! Thanks for taking a look.

tedyangx:
@tobyhs @tomlongridge I have addressed the review comments and updated the pull request. Much appreciated if you can review this. Thanks!

tobyhs:
@tedyangx : I no longer work at Bugsnag, so I won't be able to merge this.

tomlongridge:
Thank @tedyangx - we'll take a look.

tomlongridge:
Raised separate PR with these changes - #583 - to enable CI to run.

tomlongridge:
@tedyangx / @jnraine - thanks for your contributions on this 🙏 The feature has now been released on [6.13.0](https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.13.0).

PR #546, opened 2019-06-11
Specify redis gem version for Ruby 2.2

tobyhs:
## Goal

Fix the Travis build for Ruby 2.2 and Sidekiq. I also started using `Gem::Version` as that seems to be more accurate, especially when version components are more than one digit.

## Review

  - [x] Final review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,16 +3,16 @@ source "https://rubygems.org"
 ruby_version = Gem::Version.new(RUBY_VERSION)
 
 group :test, optional: true do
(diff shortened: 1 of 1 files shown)


tobyhs:
It looks like the branch build for this passed (after I retried many times). I'm now encountering the same rvm issues with the pull request build on the jruby-19mode job (where running `rvm get stable` tries to download https://github.com/rvm/rvm/archive/.tar.gz ; the branch/tag/version is missing) that doesn't appear to be related to this change (as the failure also comes up many times on a whitespace change).

PR #550, opened 2019-06-26
Make unhandled writeable

### LINKED ISSUE #549: Set the unhandled/handled state
### Description
Currently there is no easy way to set the handled/unhandled state for an exception.

In our Rails application we are catching all exceptions using `rescue_from`, to render an error json payload + send it to Bugsnag. While we do "handle" the exception, in the sense that we're catching it. The resulting status code will still be `500` and the user will see an error message.

So, from a stability perspective it wasn't handled. Thus, we'd like these errors to worsen the Bugsnag stability score. And for that to work we'd like to set the handled/unhandled state via `Bugsnag.notify`.

```ruby
rescue_from StandardError, :with => :custom_error_render
```

### Issue

We'd like more control over the handled/unhandled state. For example by making the property adjustable on the report instance.

### Environment

Library versions: 6.11.1

sandstrom:
## Goal

Make it possible to change the unhandled state, to make the stability score usable.

See more details in #549.

Closes #549

*Feel free to run with this, only adding a PR as a starting point for discussion.*

### DIFF
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -19,7 +19,7 @@ module Bugsnag
 
     CURRENT_PAYLOAD_VERSION = "4.0"
 
-    attr_reader   :unhandled
+    attr_accessor :unhandled
     attr_accessor :api_key
     attr_accessor :app_type
     attr_accessor :app_version
@@ -42,7 +42,7 @@ module Bugsnag
     # Initializes a new report from an exception.
     def initialize(exception, passed_configuration, auto_notify=false)
       @should_ignore = false
-      @unhandled = auto_notify
+      self.unhandled = auto_notify
 
       self.configuration = passed_configuration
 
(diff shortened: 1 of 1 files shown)


sandstrom:
@johnkiely1 Now that it's possible to set the unhandled state in the JS notifier, would it be possible for you to add it to the ruby notifier too?

mattdyoung:
Hi @sandstrom 
This is definitely on our radar as something we plan to support on our Ruby notifier as well in the future and we'll be looking to implement this when priorities allow. It's not as simple as the change in this PR as the way we correlate events and sessions needs to change as well to cater for whether events associated with a session change state between handled / unhandled. We'll be sure to provide updates in due course!

sandstrom:
@mattdyoung Thanks for the update & looking forward to this ability!

With the updated JS notifier we can finally use stability boards (previously they reported all our releases at 100%, despite having errors, since they were all incorrectly marked as 'unhandled'). 

Looking forward to using stability board for Ruby too!

imjoehaines:
This was implemented in https://github.com/bugsnag/bugsnag-ruby/pull/698 and released in [v6.24.0](https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.24.0)

PR #554, opened 2019-07-29
Better Rails example instructions

Cawllec:
## Goal

Updates the examples and configuration required to run Sidekiq & Que in the `rails-51` example application.

### DIFF
--- a/example/rails-51/README.md
+++ b/example/rails-51/README.md
@@ -56,6 +56,12 @@ Navigate to the `/sidekiq` page and run any of the examples using the links prov
 
 The worker code can be found in `app/workers/sidekiq_workers.rb`.
 
+To process the jobs, run Sidekiq using the following command:
+
+```shell
+bundle exec sidekiq
+```
+
 ## Que in Rails
 
 Que requires a database backend in order to queue jobs.  By default this database will be PostgreSQL although this can be changed via options as detailed in [the que documentation](https://github.com/chanks/que).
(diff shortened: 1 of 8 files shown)


example/rails-51/config/database.yml
```diff
@@ -5,6 +5,11 @@
 #   gem 'sqlite3'
 #
 default: &default
+  adapter: sqlite3
+  pool: 5
+  timeout: 5000
```
tobyhs: (GitHub won't let me comment further below) The database entry for each section using the sqlite3 adapter should ideally be a file in the db directory with an extension of sqlite3 (as opposed to just "default" in the Rails root).

example/rails-51/README.md
```diff
@@ -64,10 +70,17 @@ Once PostgreSQL is set up as detailed using [the PostgreSQL documentation](https
...
+
+```shell
+export RAILS_ENV=que
+```
+
 Ensure that the initial Que setup is complete by running:
 
 ```shell
-bundle exec bin/rake que:install
+bundle exec bin/rails generate que:install
```
tobyhs: In this example app, running db:migrate on the migration generated by running this command fails for me:
```
bugsnag-ruby/example/rails-51$ RAILS_ENV=que be rake db:migrate
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class AddQue < ActiveRecord::Migration[4.2]
```

We might need que 0.14.x for it to work (unless I'm doing something wrong).

tomlongridge: I get the same...

example/rails-51/config/database.yml
tomlongridge: We don't make use of sqlite database in our example as far as I can see, so don't think we need to update this.

tomlongridge: We don't make use of these sqlite databases in our examples as far as I can see - I don't see any benefit in updating this.

lib/bugsnag/integrations/que.rb
```diff
@@ -1,3 +1,5 @@
+require 'que'
```
tobyhs: It looks like the addition of this line causes spec/integrations/que_spec.rb to fail. If you intend to keep this line, you'll likely need to add que to the test group of the root Gemfile and possibly remove the manual Que mock in que_spec.rb.

PR #560, opened 2019-08-15
Add runtime version reporting to event and session payloads

tomlongridge:
## Goal

Augments reporting with the version of Ruby and integrated frameworks in use when an event occurred to aid diagnostics of version-specific bugs.

## Design

Added data to device.runtimeVersions.ruby as per other notifiers.

### Changed

* `configuration.rb` - adds Ruby/JRuby version to new property hash `runtime_version`
* `report.rb` - adds the collected versions into the payload
* `session_tracker.rb` - adds the collected versions into the payload
* Updated Que, Rack, Rails, Rake, Resque and Sidekiq integrations to add version

## Tests

* Augmented unit tests where appropriate
* Manually verified versions added to payload using example projects

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/middleware/rails3_request.rb
+++ b/lib/bugsnag/middleware/rails3_request.rb
@@ -32,13 +32,6 @@ module Bugsnag::Middleware
         })
 
         report.user["id"] = client_ip
-
-        # Add the rails version
-        if report.configuration.send_environment
-          report.add_tab(:environment, {
-            :railsVersion => Rails::VERSION::STRING
-          })
-        end
       end
 
       @bugsnag.call(report)
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -29,6 +29,7 @@ module Bugsnag
     attr_accessor :delivery_method
     attr_accessor :exceptions
     attr_accessor :hostname
+    attr_accessor :runtime_versions
     attr_accessor :grouping_hash
     attr_accessor :meta_data
     attr_accessor :raw_exceptions
@@ -55,6 +56,7 @@ module Bugsnag
       self.breadcrumbs = []
       self.delivery_method = configuration.delivery_method
       self.hostname = configuration.hostname
+      self.runtime_versions = configuration.runtime_versions
       self.meta_data = {}
       self.release_stage = configuration.release_stage
       self.severity = auto_notify ? "error" : "warning"
@@ -97,7 +99,8 @@ module Bugsnag
         },
         context: context,
         device: {
-          hostname: hostname
+          hostname: hostname,
+          runtimeVersions: runtime_versions
         },
         exceptions: exceptions,
         groupingHash: grouping_hash,
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -50,15 +50,15 @@ module Bugsnag
       self.raw_exceptions = generate_raw_exceptions(exception)
       self.exceptions = generate_exception_list
 
-      self.api_key = configuration.api_key.dup
-      self.app_type = configuration.app_type.dup
-      self.app_version = configuration.app_version.dup
+      self.api_key = configuration.api_key
+      self.app_type = configuration.app_type
+      self.app_version = configuration.app_version
       self.breadcrumbs = []
-      self.delivery_method = configuration.delivery_method.dup
-      self.hostname = configuration.hostname.dup
+      self.delivery_method = configuration.delivery_method
+      self.hostname = configuration.hostname
       self.runtime_versions = configuration.runtime_versions.dup
       self.meta_data = {}
-      self.release_stage = configuration.release_stage.dup
+      self.release_stage = configuration.release_stage
       self.severity = auto_notify ? "error" : "warning"
       self.severity_reason = auto_notify ? {:type => UNHANDLED_EXCEPTION} : {:type => HANDLED_EXCEPTION}
       self.user = {}
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -50,15 +50,15 @@ module Bugsnag
       self.raw_exceptions = generate_raw_exceptions(exception)
       self.exceptions = generate_exception_list
 
-      self.api_key = configuration.api_key
-      self.app_type = configuration.app_type
-      self.app_version = configuration.app_version
+      self.api_key = configuration.api_key.dup
+      self.app_type = configuration.app_type.dup
+      self.app_version = configuration.app_version.dup
       self.breadcrumbs = []
-      self.delivery_method = configuration.delivery_method
-      self.hostname = configuration.hostname
-      self.runtime_versions = configuration.runtime_versions
+      self.delivery_method = configuration.delivery_method.dup
+      self.hostname = configuration.hostname.dup
+      self.runtime_versions = configuration.runtime_versions.dup
       self.meta_data = {}
-      self.release_stage = configuration.release_stage
+      self.release_stage = configuration.release_stage.dup
       self.severity = auto_notify ? "error" : "warning"
       self.severity_reason = auto_notify ? {:type => UNHANDLED_EXCEPTION} : {:type => HANDLED_EXCEPTION}
       self.user = {}
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -1,4 +1,6 @@
 # encoding: utf-8
+require 'json'
+require 'socket'
 require 'spec_helper'
 
 describe Bugsnag::Configuration do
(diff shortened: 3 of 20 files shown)


tobyhs:
Should this target the `next` branch instead of `master`?

spec/report_spec.rb
```diff
@@ -1267,4 +1267,16 @@ def gloops
       }
     end
   end
+
+  it 'should include device data when notify is called' do
```
tobyhs: I prefer avoiding "should" in doc strings (https://github.com/rubocop-hq/rspec-style-guide#should-in-example-docstrings) to avoid the feeling of repeating "should" in most doc strings.

lib/bugsnag/report.rb
```diff
@@ -55,6 +56,7 @@ def initialize(exception, passed_configuration, auto_notify=false)
       self.breadcrumbs = []
       self.delivery_method = configuration.delivery_method
       self.hostname = configuration.hostname
+      self.runtime_versions = configuration.runtime_versions
```
tobyhs: This should probably be assigned a duplicate (`configuration.runtime_versions.dup`) instead, especially if it is mutable via the `Report` that is yielded in `Bugsnag.notify` (because if for some reason I feel like writing something like `report.runtime_versions['something'] = '1.2.3'`, I shouldn't expect it to change the Bugsnag configuration).

spec/configuration_spec.rb
```diff
@@ -1,4 +1,6 @@
 # encoding: utf-8
+require 'json'
```
tobyhs: I'm not sure why this was added as I don't see any explicit uses of the json gem/library in this file.

spec/integrations/rake_spec.rb
```diff
@@ -64,6 +65,8 @@
       result = request()
       expect(result["events"][0]["metaData"]["rake_task"]).not_to be_nil
       expect(result["events"][0]["metaData"]["rake_task"]["name"]).to eq("test:crash")
+      expect(result["events"][0]["app"]["type"]).to eq("rake")
+      expect(result["events"][0]["device"]["runtimeVersions"]["rake"]).to match(/\d+\.\d+\.\d+/)
```
tobyhs: Anchor the regular expression (use `\A` in the beginning, although using `\z` at the end might be too brittle in cases where the version string is a prerelease or has 4 components).

spec/integrations/sidekiq_spec.rb
```diff
@@ -29,6 +30,8 @@ def perform(value)
         expect(event["metaData"]["sidekiq"]["msg"]["args"]).to eq([-0])
         expect(event["metaData"]["sidekiq"]["msg"]["queue"]).to eq("default")
         expect(event["severity"]).to eq("error")
+        expect(event["app"]["type"]).to eq("sidekiq")
+        expect(event["device"]["runtimeVersions"]["sidekiq"]).to eq(::Sidekiq::VERSION)
```
tobyhs: This should also have some version string sanity check as Sidekiq isn't faked/mocked

lib/bugsnag/report.rb
tomlongridge: Seems reasonable - I've applied this to the other fields taken from `configuration` and assigned to object variables in `init`.

lib/bugsnag/middleware/rails3_request.rb
```diff
@@ -32,13 +32,6 @@ def call(report)
         })
 
         report.user["id"] = client_ip
-
-        # Add the rails version
-        if report.configuration.send_environment
-          report.add_tab(:environment, {
-            :railsVersion => Rails::VERSION::STRING
```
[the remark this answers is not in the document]
tomlongridge: Created and branch (and PR to follow) for this.

lib/bugsnag/report.rb
```diff
@@ -49,14 +50,15 @@ def initialize(exception, passed_configuration, auto_notify=false)
       self.raw_exceptions = generate_raw_exceptions(exception)
       self.exceptions = generate_exception_list
 
-      self.api_key = configuration.api_key
-      self.app_type = configuration.app_type
-      self.app_version = configuration.app_version
+      self.api_key = configuration.api_key.dup
+      self.app_type = configuration.app_type.dup
+      self.app_version = configuration.app_version.dup
```
tobyhs: If I apply the following diff, I can reproduce and inspect the failing test (by running `bundle exec rspec spec/integrations/rake_spec.rb` under Ruby 2.3.8):
```diff
--- a/spec/integrations/rake_spec.rb
+++ b/spec/integrations/rake_spec.rb
@@ -59,7 +59,7 @@ describe "Bugsnag Rake integration" do
       ENV['BUGSNAG_TEST_SERVER_PORT'] = server.config[:Port].to_s
       task_fixtures_path = File.join(File.dirname(__FILE__), '../fixtures', 'tasks')
       Dir.chdir(task_fixtures_path) do
-        system("bundle exec rake test:crash > /dev/null 2>&1")
+        system("bundle exec rake test:crash")
       end

       result = request()
```

The relevant resulting output is:
```
rake aborted!
TypeError: can't dup NilClass
/Users/toby/work/bugsnag-ruby/lib/bugsnag/report.rb:55:in `dup'
/Users/toby/work/bugsnag-ruby/lib/bugsnag/report.rb:55:in `initialize'
/Users/toby/work/bugsnag-ruby/lib/bugsnag.rb:70:in `new'
/Users/toby/work/bugsnag-ruby/lib/bugsnag.rb:70:in `notify'
/Users/toby/work/bugsnag-ruby/lib/bugsnag/integrations/rake.rb:21:in `rescue in execute'
/Users/toby/work/bugsnag-ruby/lib/bugsnag/integrations/rake.rb:31:in `execute'
/Users/toby/work/bugsnag-ruby/bundle/ruby/2.3.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/Users/toby/.rbenv/versions/2.3.8/bin/bundle:22:in `load'
/Users/toby/.rbenv/versions/2.3.8/bin/bundle:22:in `<main>'
```

In Ruby 2.3 and lower, calling `dup` on `nil` fails. Since this pull request already has more than 20 files changed, I would suggest just using `dup` on your change (`configuration.runtime_versions.dup`) for now to avoid possible issues.

spec/configuration_spec.rb
tobyhs: I'm not sure why this comment was marked as resolved. This file should require `spec_helper` like before because spec_helper.rb contains global setup/configuration for tests. Additionally, it isn't clear why to me why you're requiring json in this file.

lib/bugsnag/report.rb
tomlongridge: Thanks @tobyhs!

spec/configuration_spec.rb
tomlongridge: Apologies - I applied the change the wrong way round for some reason! Now removed `json` and restored `spec_helper`.

spec/integrations/rake_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'webrick'
 require 'spec_helper'
 require 'json'
+require 'rake'
```
tobyhs: This require is no longer needed, unless you want to add an additional expectation between lines 68-69 (something like `expect(result["events"][0]["device"]["runtimeVersions"]["rake"]).to eq(::Rake::VERSION)`)

spec/integrations/sidekiq_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'sidekiq'
```
tobyhs: This require appears to be unnecessary.

spec/integrations/sidekiq_spec.rb
```diff
@@ -29,6 +30,8 @@ def perform(value)
         expect(event["metaData"]["sidekiq"]["msg"]["args"]).to eq([-0])
         expect(event["metaData"]["sidekiq"]["msg"]["queue"]).to eq("default")
         expect(event["severity"]).to eq("error")
+        expect(event["app"]["type"]).to eq("sidekiq")
+        expect(event["device"]["runtimeVersions"]["sidekiq"]).to match(/\A\d+\.\d+\.\d+/)
```
tobyhs: It turns out I was wrong at https://github.com/bugsnag/bugsnag-ruby/pull/560#discussion_r314879567.

The version is stubbed via `stub_const` on line 18, so this should probably be something like `.to eq('2.0.0')` instead.

spec/integrations/rake_spec.rb
tomlongridge: We removed that in place of a regex, so I'll remove the require.

PR #565, opened 2019-09-02
fix(rails): fix nil dereference in SQL Active Record `bind` key

### LINKED ISSUE #555: NoMethodError after updating to 6.11.1
### Description
I’m trying to update the bugsnag gem from 6.10.0 to 6.11.1 for a rails application and the following error started to pop up when running the test suite. I'm able to run 6.10.0 without any issues.

```
undefined method `each_with_object' for nil:NilClass (NoMethodError)
/gems/bugsnag-6.11.1/lib/bugsnag/integrations/railtie.rb:82:in `block in event_subscription'
```

Might be related to using `mysql2` as [discussed here](https://github.com/bugsnag/bugsnag-ruby/pull/521#discussion_r245392743).

Configuration for bugsnag is pretty minimal.

```
Bugsnag.configure do |config|
  config.api_key =
  config.meta_data_filters +=
  config.release_stage ||= ENV['RAILS_ENV'] || 'development'
  config.ignore_classes += [SignalException, SystemExit]
end
```

### Environment
* Rails 3.2.22
* Bugsnag 6.11.1
* mysql2 0.3.21

tomlongridge:
## Goal

Older versions of `mysql2` adapter do not contain the `bind` key in the Active Record notification.

### Changed

* `railtie.rb` - added extra guard for key not being present in notification

## Tests

I was unable to get a Rails instance with `mysql2` v0.3.x working to reproduce this issue. However I have run it with later versions to ensure that we pick up the `bind` key correctly still (this is also checked on MazeRunner tests).

### Linked issues

Fixes #555

## Review

  - [ ] Final review


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+## TBC
+
+### Fixes
+
+* Fix nil dereference in SQL Active Record `bind` key.
```
tobyhs: `:binds` instead of `bind`, although I would probably reword this to something like "Account for missing `:binds` key in `sql.active_record` ActiveSupport notifications"

PR #569, opened 2019-10-16
adds deep string filter for nested keys/params to cleaner.rb

### LINKED ISSUE #445: Filter nested parameter
### Expected behavior
From Rails 5.0 [we can filter nested parameter](http://api.rubyonrails.org/classes/ActionDispatch/Http/FilterParameters.html) with

```ruby
env["action_dispatch.parameter_filter"] = [ "credit_card.code" ]
=> replaces { credit_card: {code: "xxxx"} } with "[FILTERED]", does not
change { file: { code: "xxxx"} }
```

```ruby
Rails.application.config.filter_parameters += ["form.xxx"]

param_filter = ActionDispatch::Http::ParameterFilter.new(Rails.application.config.filter_parameters)
param_filter.filter({ "form" => { "xxx" => "123456" } })
=> {"form"=>{"xxx"=>"[FILTERED]"}}
```

### Observed behavior

```ruby
Bugsnag.configuration.meta_data_filters
=> #<Set: {/authorization/i,
 /cookie/i,
 /password/i,
 /secret/i,
 "rack.request.form_vars",
 /\Apassword\z/,
 /\Aform.xxx\z/}>

Bugsnag::Cleaner.new(Bugsnag.configuration.meta_data_filters).clean_object({ "form" => { "xxx" => "123456" } })
=> {"form"=>{"xxx"=>"123456"}}
```

### Version
Bugsnag 6.6.4
Rails 5.1.5

Craig-Kelly:
## Goal

Correctly detect presence of deep String filters to allow them to be used in isolation. As described in [config.meta_data_filters](https://docs.bugsnag.com/platforms/ruby/rails/configuration-options/#meta_data_filters)

## Context

If deep filters are detected, deep string matching works as expected. However, only deep Regex filters are detected as well, so deep filtering is skipped if only Strings are provided.

## Design

Add `deep_filter?` method to check for dot separated Regexp and String filters

## Changeset

`cleaner.rb`

Removed: `@deep_filters` from `initialize(filters)`

Added: 
Detection of String filters, instead of just Regex filters

## Tests

Added tests for deep string filters in isolation

### Alternative Approaches

Instead of memoizing, could have kept setup in initialize, but still delegate to private method to scan `@filters`

### Linked issues

Fixes #445 

## Review

  - [ ] Final review

### DIFF
--- a/lib/bugsnag/cleaner.rb
+++ b/lib/bugsnag/cleaner.rb
@@ -7,10 +7,11 @@ module Bugsnag
     RECURSION = '[RECURSION]'.freeze
     OBJECT = '[OBJECT]'.freeze
     RAISED = '[RAISED]'.freeze
+    ESCAPED_PERIOD = '\.'.freeze
+    PERIOD = '.'.freeze
 
     def initialize(filters)
       @filters = Array(filters)
-      @deep_filters = @filters.any? {|f| f.kind_of?(Regexp) && f.to_s.include?("\\.".freeze) }
     end
 
     def clean_object(obj)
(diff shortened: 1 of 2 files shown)


mattdyoung:
Hi @Craig-Kelly 
Thanks for the PR. It looks like this should make the behaviour more consistent so we'll take a look when we have the bandwidth to perform the necessary review and testing before merging.

Until then I presume this can be worked around by using a regex?

Thanks!

imjoehaines:
Thanks for the PR @Craig-Kelly!

The behaviour for Regexps exists because we automatically import `filter_parameters` in Rails apps, which works correctly for String filters but some Regexps won't match when it seems like they should. For ease of integration, we do this deep matching to ensure that Rails `filter_parameters` behave as expected when imported into Bugsnag's `metadata_filters`

We've been discussing this internally and think it's best if we handled this specifically in the Rails integration and removed the "deep filter" support from the generic cleaner. This is because the behaviour can be unexpected/confusing, lead to false positives and there's a lot of complexity if we fully supported these "dot paths"

Hopefully that makes sense, thanks again for the PR 🙂

PR #570, opened 2019-10-21
Error loading with Que integration due to capitalised VERSION

tommeier:
## Goal

Initialising bugsnag fails with `Que` due to titlecase `Version` expected, when `Que` uses uppercase `VERSION`

  * https://github.com/que-rb/que/blob/master/lib/que/version.rb#L4

Screen Shot 2019-10-22 at 8 40 16 am


tomlongridge:
Thanks for the report @tommeier. It looks like the capitalisation changed in the 1.x (beta) branch so we'll need a way to cover both:

https://github.com/que-rb/que/blob/0.x/lib/que/version.rb#L4

tomlongridge:
Fixed in #572

PR #574, opened 2019-11-05
Move tests to buildkite, adding e2e tests

Cawllec:
## Goal

Moves all tests to Buildkite.
Adds a full set of e2e tests covering all supported ruby and rails versions


features/steps/ruby_notifier_steps.rb
```diff
@@ -1,31 +1,79 @@
...
+
+    And the payload field "notifier.name" equals "#{notifier_name}"
+    And the payload field "notifier.url" is not null
+    And the payload field "notifier.version" is not null
+    And the payload field "events" is a non-empty array
+
+    And each element in payload field "events" has "severity"
+    And each element in payload field "events" has "severityReason.type"
+    And each element in payload field "events" has "unhandled"
+    And each element in payload field "events" has "exceptions"
```
tomlongridge: Remove?

features/steps/ruby_notifier_steps.rb
```diff
@@ -1,31 +1,79 @@
...
-    the "#{element}" of stack frame #{frame_index} equals #{value}
+    When I open the URL "http://rails#{rails_version}:3000#{route}"
   }
 end
 
-Then(/^the total sessionStarted count equals (\d+)(?: for request (\d+))?$/) do |value, request_index|
-  session_counts = read_key_path(find_request(request_index)[:body], "sessionCounts")
-  total_count = session_counts.inject(0) { |count, session| count += session["sessionsStarted"] }
-  assert_equal(value, total_count)
+Then("the payload field {string} matches the appropriate handled JSON fixture") do |field|
```
tomlongridge: The step text is too generic for its purpose, which relates only Sidekiq.

```suggestion
Then("the payload field {string} matches the appropriate Sidekiq handled payload") do |field|
```

features/steps/ruby_notifier_steps.rb
```diff
@@ -1,31 +1,79 @@
...
+  steps %Q{
+    And the payload field "#{field}" matches the JSON fixture in "features/fixtures/sidekiq/payloads/unhandled_metadata_ca_#{created_at_present}.json"
+  }
+end
+
+Then("the event contains a breadcrumb matching the JSON fixture in {string}") do |json_fixture|
+  breadcrumbs = read_key_path(Server.current_request[:body], "events.0.breadcrumbs")
+  expected = JSON.parse(open(json_fixture, &:read))
+  match = breadcrumbs.any? { |breadcrumb| value_compare(expected, breadcrumb).equal? }
+  assert(match, "No breadcrumbs in the event matched the given breadcrumb")
```
tomlongridge: Should this be moved to Maze Runner?

features/steps/ruby_notifier_steps.rb
Cawllec: So this step is a response to the [maze-runner version of the step](https://github.com/bugsnag/maze-runner/blob/v2/lib/features/steps/error_reporting_steps.rb#L8) which includes the check `the payload contains the payloadVersion "#{payload_version}"` which will always fail in the Ruby notifier as it does not include the payload version in the payload, only the header.

Upon discussion it was unclear whether this check was necessary for the running of the notifier, and whether or not it could be removed, but no clear action was determined.

features/steps/ruby_notifier_steps.rb
Cawllec: PR present [here](https://github.com/bugsnag/maze-runner/pull/73)

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,411 @@
...
+          image-repository: 855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby
+          cache-from: ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby${BRANCH_NAME}
+      - docker-compose#v2.6.0:
+          push:
+            - ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby${BRANCH_NAME}
+            - ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby-latest
+
+  - wait
+
+  - label: ':ruby: Ruby 1.9.3 unit tests'
```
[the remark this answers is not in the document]
Cawllec: I'll add a table to the notifier docs page, with the CI setup.  I've updated that with an explanation of what's running where.

Issue #478, opened 2018-07-30
NoMethodError · undefined method `call' for nil:NilClass


tibbon:
After updating the bugsnag gem from 6.7.3 to 6.8.0 the following error started popping up in Bugsnag as soon as our Rails server started, and then periodically afterward. As you can see, there isn't terribly much information to go on. The error however did come in pairs, and always appeared twice in the Bugsnag UI. Nothing of note appeared in our logging service. 

I don't believe we have anything special happening in our bugsnag config, but if there are specific questions I can answer them. 

```
NoMethodError · undefined method `call' for nil:NilClass 
Did you mean? caller 
unknown file
unknown method
```

I stepped through the commits between 6.7.3 and 6.8.0 and narrowed it down between two listed below. Unfortunately, I don't see anything obvious that would trigger this, but the `os` gem is a new addition. 

WORKS: https://github.com/bugsnag/bugsnag-ruby/commit/327dd014cba2d35c2c0e6ea5de78160cc0488b7c
ERROR: https://github.com/bugsnag/bugsnag-ruby/commit/88f04913359578ce848a513b14644083d1fee7c3

kattrali:
Thanks for the report, @tibbon, we are looking into it but haven't turned anything up yet. Can you share how you are configuring Bugsnag, such as the contents of your `config/initializers/bugsnag.rb` (minus the API key)? You can also send it support@bugsnag.com as well if there are sensitive contents or additional setup information that you wouldn't want to discuss on a public forum.

The commits in question only affect integration test gemsets, so there might be something else going on here.

etlund:
I just upgraded Bugsnag to 6.8.0 in my Rails app and have started seeing these "undefined method 'call' for nil:NilClass" errors too. Each one corresponds to a WARN in my production.log:
`[Bugsnag]: Waiting for 1 outstanding request(s)`

My initializer is pretty minimal:
````
Bugsnag.configure do |config|
  config.api_key = 
  config.release_stage = Rails.env
  config.notify_release_stages = %w(production staging)
end
````

etlund:
Hi!  I'm running Rails 5.1.6.  I think I upgraded from 5.25.2.

Thanks for checking it out.

Cawllec:
I'm attempting to reproduce and track down the source of this issue.  It would be useful if you could let me know (securely at [support@bugsnag.com](support@bugsnag.com) if necessary):
- Gems used outside of the standard Rails 5 gemset
[...]

PR #575, opened 2019-11-08
Fix crash no backtrace

mathieujobin:
## Goal

Anyone else having this crash within the gem?
image

I must have an uncommon type of exception.

I did not have any problem with bugstand 6.6.4
then upgraded to 6.12.2, rollback to 6.11 for testing, but its the same
made this quick patch

Why was this approach to the goal used?

I thought I could simple check if the backtrace was present, but it still crash with the same error
so I am now testing a rescue.
as I am writing this, I have not yet deployed this rescue version.

### Changed

added rescue

## Tests

i have not planned on adding any tests for this myself.
mostly reporting the problem in a form of PR in order to fix it for myself.
will likely test deploy it, at least I can find out what exception is being raised.
at this point, I dont know why it started happening.

### Alternative Approaches

moving to another service?
rolling back to 6.6.4?

### Outstanding Questions

need to locate if there is another cause

### Linked issues

n/a

## Review

  - [ ] Final review

### DIFF
--- a/lib/bugsnag/report.rb
+++ b/lib/bugsnag/report.rb
@@ -192,9 +192,15 @@ module Bugsnag
         {
           errorClass: error_class(exception),
           message: exception.message,
-          stacktrace: (exception.backtrace ? Stacktrace.new(exception.backtrace, configuration).to_a : nil)
+          stacktrace: trace_object(exception)
         }
       end
+    end
+    
+    def trace_object(exception)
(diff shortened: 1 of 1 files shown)


snmaynard:
I believe this might be #478 which is an issue with passenger. Could that explain what you are seeing?

mathieujobin:
Thank you @snmaynard I will try that, that makes sense.
my little patch does not work anyway,

closing...

PR #582, opened 2020-01-27
fix: Resolve Ruby deprecation warning (in Ruby 2.7)/error (in 3.0)

### LINKED ISSUE #580: cleaner.rb:65: warning: Using the last argument as keyword parameters is deprecated
### Description
Run on Ruby 2.7

### Issue
Deprecation warnings are sent to stdout:

```
/vendor/bundle/ruby/2.7.0/gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:65: warning: Using the last argument as keyword parameters is deprecated
```

### Environment

Library versions: bugsnag 6.12.2
Ruby: 2.7

- OS / OS version: Ubuntu
- production

See https://piechowski.io/post/last-arg-keyword-deprecated-ruby-2-7/

twometresteve:
## Goal

Resolve Ruby warning generated by Ruby 2.7 (and that become errors in v3).

## Tests

Tested using existing Unit and Component tests.

### Linked issues

Fixes #580

### DIFF
--- a/lib/bugsnag/cleaner.rb
+++ b/lib/bugsnag/cleaner.rb
@@ -2,7 +2,6 @@ require 'uri'
 
 module Bugsnag
   class Cleaner
-    ENCODING_OPTIONS = {:invalid => :replace, :undef => :replace}.freeze
     FILTERED = '[FILTERED]'.freeze
     RECURSION = '[RECURSION]'.freeze
     OBJECT = '[OBJECT]'.freeze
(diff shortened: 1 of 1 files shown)


Cawllec:
Was this previously picked up in our tests, like failing CI at all?  If not it might be good to add deprecation checks to our "latest" test versions (currently 2.7)

twometresteve:
> Was this previously picked up in our tests, like failing CI at all? If not it might be good to add deprecation checks to our "latest" test versions (currently 2.7)

It didn't cause tests to fail, we just happened to notice (as did at least one user).  I like the idea of checking for things like this in CI, but I'll raise a separate ticket to address that I think.

Issue #577, opened 2019-12-13
`to_s': stack level too deep (SystemStackError)


igorescobar:
### Description
We are having this error on our bugsnag account every since we installed this gem to integrate with bugsnag. Is this known? The information we have it's not really helpful so I thought about reporting it to see if this is really happening from the gem itself or not.

Btw: This is happening only for our delayed_jobs

### Issue
```
gems/activesupport-6.0.1/lib/active_support/core_ext/numeric/conversions.rb:107:in `to_s': stack level too deep (SystemStackError)
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:100:in `filters_match?'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:115:in `filters_match_deeply?'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:33:in `block in traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:32:in `each'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:32:in `traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:36:in `block in traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:32:in `each'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:32:in `traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:41:in `block in traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:41:in `map'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:41:in `traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:36:in `block in traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:32:in `each'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:32:in `traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:41:in `block in traverse_object'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:41:in `map'
    from gems/bugsnag-6.12.2/lib/bugsnag/cleaner.rb:41:in `traverse_object'
[...]

PR #591, opened 2020-05-01
Guard against exceptions in to_s when cleaning objects

imjoehaines:
## Goal

When sending notifications we run through a "cleaning" stage which does things like filter out any keys that we shouldn't send in the error report (e.g. passwords). This calls `to_s` in a couple of different places, which can cause issues as objects are free to implement any logic and can therefore throw exceptions or infinitely recurse, which means we end up crashing while trying to cleanup the object

In this PR we'll now recover from exceptions and stack overflows by replacing the values with one of our existing replacement strings:

1. `[RECURSION]` if we get a `SystemStackError` when stringifying a plain object
1. `[FILTERED]` if we get an exception when attempting to filter keys in a hash map — if an exception happens we can't tell if we should filter the key, so we have to assume we need to

This means the error report will have less information that intended, but this should be rare and if it does happen then the alternative is no report at all

### Changed

- Rescue `SystemStackError` as well as other exceptions when calling `to_s` on an object
- Rescue `StandardError` and `SystemStackError` when calling `to_s` on a key when filtering a hash map

## Tests

Tests have been added for:
- hash keys that throw during `to_s`
- hash values that throw during `to_s`
- custom objects that infinitely recurse during `to_s`

### Linked issues

Related to #577

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/cleaner.rb
+++ b/lib/bugsnag/cleaner.rb
@@ -43,7 +43,13 @@ module Bugsnag
       when String
         clean_string(obj)
       else
-        str = obj.to_s rescue RAISED
+        # guard against objects that raise or blow the stack when stringified
+        begin
+          str = obj.to_s rescue RAISED
+        rescue SystemStackError
+          str = RECURSION
+        end
+
         # avoid leaking potentially sensitive data from objects' #inspect output
         if str =~ /#<.*>/
           OBJECT
@@ -96,7 +102,13 @@ module Bugsnag
     private
 
     def filters_match?(key)
-      str = key.to_s
+      # If we can't stringify this key, we assume it needs to be filtered to
+      # avoid leaking things we shouldn't
+      begin
+        str = key.to_s
+      rescue StandardError, SystemStackError
+        return true
+      end
 
       @filters.any? do |f|
         case f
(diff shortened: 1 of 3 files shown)


lib/bugsnag/cleaner.rb
```diff
@@ -43,7 +43,13 @@ def traverse_object(obj, seen, scope)
       when String
         clean_string(obj)
       else
-        str = obj.to_s rescue RAISED
+        # guard against objects that raise or blow the stack when stringified
+        begin
+          str = obj.to_s rescue RAISED
```
tomlongridge, line 48: Does the `rescue RAISED` not catch all errors already?

imjoehaines: It catches all errors that inherit from `StandardError`, but `SystemStackError` doesn't so it won't be caught here without the separate `rescue`

I can make another `rescue` block for `StandardError` explicitly if it's confusing; something like:

```ruby
begin
  str = obj.to_s
rescue StandardError
  str = RAISED
rescue SystemStackError
  str = RECURSION
end
```

tomlongridge: Cool - yes, that would help (me!)

imjoehaines: Made the change in a1ad37e :)

PR #592, opened 2020-05-05
Replace keys with safe values when to_s fails

imjoehaines:
## Goal

This PR fixes a issue with https://github.com/bugsnag/bugsnag-ruby/pull/591 where we filter the value of a hash if its key raises, but we leave the key alone. That means that code later on can cause the same issue we were trying to fix — if they run `to_s` on the key then it will raise or infinitely recurse

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/cleaner.rb
+++ b/lib/bugsnag/cleaner.rb
@@ -29,10 +29,19 @@ module Bugsnag
       when Hash
         clean_hash = {}
         obj.each do |k,v|
-          if filters_match_deeply?(k, scope)
-            clean_hash[k] = FILTERED
-          else
-            clean_hash[k] = traverse_object(v, seen, [scope, k].compact.join('.'))
+          begin
+            if filters_match_deeply?(k, scope)
+              clean_hash[k] = FILTERED
+            else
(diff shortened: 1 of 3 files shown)


lib/bugsnag/cleaner.rb
```diff
@@ -29,10 +29,19 @@ def traverse_object(obj, seen, scope)
...
+            else
+              clean_hash[k] = traverse_object(v, seen, [scope, k].compact.join('.'))
+            end
+          # If we get an error here, we assume the key needs to be filtered
+          # to avoid leaking things we shouldn't. We also remove the key itself
+          # because it may cause issues later e.g. when being converted to JSON
+          rescue StandardError
+            clean_hash[RAISED] = FILTERED
+          rescue SystemStackError
+            clean_hash[RECURSION] = FILTERED
```
imjoehaines, line 44: Reusing the same key here does mean we'd overwrite it if more than one key throws, but I think that's unlikely to happen realistically and I don't think you'd get any useful information out of `[RAISED 1]... [RAISED 2]` if we made them unique

PR #597, opened 2020-05-15
Add discard_classes option to support strings & regular expressions and deprecate ignore_classes

### LINKED ISSUE #589: config.ignore_classes should deal in strings
### Description
`config.ignore_classes` should take an array of strings instead of constants (or either?)

### Issue
As we add error classes to the `ignore_classes` we're realizing that we're inadvertently loading those classes that don't need to be loaded. It would be good if `ignore_classes` can take in an array of class names as strings since this comparison is done at the time of error anyways and the raised error can just be converted to a string for comparison.

imjoehaines:
## Goal

The `ignore_classes` option currently supports `Class` constants and `Proc`. Rails has recently deprecated autoloading in initialization (where Bugsnag is usually configured) which makes `Class` constants unusable. Support for `Proc` is powerful but un-ergonomic when simply trying to ignore an error by class name

In order to help bring the Ruby notifier in-line with other notifiers I've added `discard_classes` as a new option and deprecated the existing `ignore_classes` option. `discard_classes` works similarly to `ignore_classes`, with these differences:

1. It accepts `String` or `Regexp` rather than `Class` or `Proc`
2. It doesn't walk the ancestor chain

### Linked issues

Fixes #589

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -127,8 +127,10 @@ module Bugsnag
 
       # SystemExit and SignalException are common Exception types seen with
       # successful exits and are not automatically reported to Bugsnag
+      # TODO move these defaults into `discard_classes` when `ignore_classes`
+      #      is removed
       self.ignore_classes = Set.new([SystemExit, SignalException])
-      self.discard_classes = Set.new(["SystemExit", "SignalException"])
+      self.discard_classes = Set.new([])
 
       # Read the API key from the environment
       self.api_key = ENV["BUGSNAG_API_KEY"]
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -3,6 +3,7 @@ require "socket"
 require "logger"
 require "bugsnag/middleware_stack"
 require "bugsnag/middleware/callbacks"
+require "bugsnag/middleware/discard_error_class"
 require "bugsnag/middleware/exception_meta_data"
 require "bugsnag/middleware/ignore_error_class"
 require "bugsnag/middleware/suggestion_data"
@@ -36,6 +37,7 @@ module Bugsnag
     attr_accessor :hostname
     attr_accessor :runtime_versions
     attr_accessor :ignore_classes
+    attr_accessor :discard_classes
     attr_accessor :auto_capture_sessions
 
     ##
@@ -123,6 +125,7 @@ module Bugsnag
       # SystemExit and SignalException are common Exception types seen with
       # successful exits and are not automatically reported to Bugsnag
       self.ignore_classes = Set.new([SystemExit, SignalException])
+      self.discard_classes = Set.new(["SystemExit", "SignalException"])
 
       # Read the API key from the environment
       self.api_key = ENV["BUGSNAG_API_KEY"]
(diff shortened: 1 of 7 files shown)


lib/bugsnag/middleware/ignore_error_class.rb
```diff
@@ -10,9 +10,16 @@ def initialize(bugsnag)
     def call(report)
       ignore_error_class = report.raw_exceptions.any? do |ex|
         ancestor_chain = ex.class.ancestors.select { |ancestor| ancestor.is_a?(Class) }.to_set
+        ancestor_chain_strings = ancestor_chain.map(&:to_s)
```
tomlongridge, line 13: Having seen this in the test, I wonder whether matching strings and regexes against ancestors is confusing. I can see if you pass an actual class in then matching subclasses sounds reasonable, but with strings and particularly regexes I think we should probably just match against the exception class. But I guess it could be argued that it would be confusing to have two different behaviours...

imjoehaines: Yeah I can see the arguments both ways

I went for this way because the migration path to fix the Rails deprecation is as easy as wrapping the current values in quotes

If we do an exact match then you'd need to find out each concrete exception class you want to ignore if you were relying on ignoring an ancestor using a class constant

Then because of that, `Regexp` being the only value handled differently would feel weird, though I can see how it might be very unexpected to match against every ancestor

lib/bugsnag/middleware/discard_error_class.rb
```diff
@@ -0,0 +1,30 @@
...
+    ##
+    # @param middleware [#call] The next middleware to call
+    def initialize(middleware)
+      @middleware = middleware
+    end
+
+    ##
+    # @param report [Report]
+    def call(report)
+      ignore_error_class = report.raw_exceptions.any? do |ex|
```
tomlongridge, line 15: I know it's just a local variable name, but perhaps better refer to it as "discarding" rather than "ignoring" here?

lib/bugsnag/configuration.rb
```diff
@@ -123,6 +128,7 @@ def initialize
       # SystemExit and SignalException are common Exception types seen with
       # successful exits and are not automatically reported to Bugsnag
       self.ignore_classes = Set.new([SystemExit, SignalException])
+      self.discard_classes = Set.new(["SystemExit", "SignalException"])
```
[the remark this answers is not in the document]
imjoehaines: Yes, so long as we remember to move them :p

How do we document it in that case? Currently we [list the default values in `ignore_classes`](https://docs.bugsnag.com/platforms/ruby/other/configuration-options/#ignore_classes) but if we add a deprecation notice and point people to `discard_classes`, I think we'd want to list the defaults there

Alternatively would you be ok with building the `discard_classes` from `ignore_classes` to remove the repetition? i.e. something like:

```ruby
self.ignore_classes = Set.new([SystemExit, SignalException])
self.discard_classes = self.ignore_classes.map(:&to_s)
```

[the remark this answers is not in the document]
imjoehaines: Removed the defaults in 4cced91b54b891f598b8a35d1196c439508a11ad

pic:
support of `Proc` would still be useful so in the proc I could check `is_a?` and still be able to group the exceptions that I want to ignore with a common ancestor. Is there any chance that it will be added again?

snmaynard:
You should be able to use the newly added `on_error` Procs to handle that

ngan:
So is the suggested way to use `on_error` for walking ancestors? Or continue using adding a `Proc` to `ignore_classes`?

imjoehaines:
> So is the suggested way to use `on_error` for walking ancestors? Or continue using adding a `Proc` to `ignore_classes`?

Yeah I'd recommend using `on_error` for that. If you're already using `ignore_classes` then it will continue to work for now, but you'll need to migrate eventually

Issue #481, opened 2018-08-06
Error reporting causes performance issues


isaacseymour:
We've noticed a couple of situations where the error rate is just high enough (e.g. someone deploys a broken version of a page), that the load caused by the Bugsnag middleware is high enough to cause performance issues on unaffected pages. 

Having a dig through the code, it seems like there's a lot of CPU time spent preparing the notification in the main request-handling thread, before the actual HTTP work is offloaded to a background thread.

Would you be open to a PR moving that work to the background thread?

snmaynard:
I think that the reason that it isn't done there is that we need to block for a certain amount of time to ensure we gather the requisite context from the thread locals etc before moving on. I think that we should raise a ticket internally to look at refactoring some of the library to improve performance.

Are there any areas you see as being especially bad for you? We'd have to look at what we can remove or how we can serialize information before offloading work to the background thread.

snmaynard:
We have also raised an internal ticket to audit the performance and see if there are any quick wins in our test cases.

gigr:
Has there been any movement on this? We have also been bitten by this issue. It cropped up recently in a case where some code inadvertently sends events to Bugsnag in a loop and can take 10–15 seconds to send a few dozen errors. We see heavy CPU usage during these times. We've started to mitigate this by sending events to Bugsnag in background jobs, but it's not ideal.

xander-jones:
Hi @gigr, we haven't identified any performance improvements in this area yet.

Do you know what is causing the slowdown in your case? Have you profiled which paths are the worst offenders for this? 

Additionally, if it is possible for you to share a reproduction case, this would help us target performance upgrades.

imjoehaines:
We've shipped a number of performance improvements in recent releases; let us know if these don't seem to have helped

PR #604, opened 2020-07-15
Code extraction performance improvements

imjoehaines:
## Goal

This PR improves the performance of extracting code from files in a stacktrace

The performance improvements are especially relevant when a stacktrace has the same file multiple times (e.g. when chaining private methods) and when the files are long

Exactly how much better this is is impossible to say without looking at a specific stacktrace — it can be more than a 3x speedup if it mostly consists of the same few files and they are very long, but every stacktrace is different 🌈. I didn't observe any noticable slow down when a stackrace contained exclusively different files, but the new code does involve a bit more work in the worst case scenario

### Added

- Code extraction is now handled by a `CodeExtractor` class. This exists to keep track of which files in the stacktrace we need to get code from (we will sometimes remove files from the processed stacktrace) and so that it can do a single pass per file, extracting all the regions of code that it needs in one go
- The `Stacktrace`class is now a module as it was always basically one method and was immediately converted to an array when it was called
- A bunch of new tests and fixtures for code extraction and some untested parts of the stacktrace processing

### Linked issues

Related to #481

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/stacktrace.rb
+++ b/lib/bugsnag/stacktrace.rb
@@ -17,7 +17,7 @@ module Bugsnag
     #
     # rubocop:todo Metrics/CyclomaticComplexity
     def self.process(backtrace, configuration)
-      code_extractor = CodeExtractor.new
+      code_extractor = CodeExtractor.new(configuration)
 
       backtrace = caller if !backtrace || backtrace.empty?
 
--- a/lib/bugsnag/stacktrace.rb
+++ b/lib/bugsnag/stacktrace.rb
@@ -60,7 +60,7 @@ module Bugsnag
         # Add a method if we have it
         trace_hash[:method] = method if method && (method =~ /^__bind/).nil?
 
-        # TODO test this
+        # TODO: test this
         return nil unless trace_hash[:file] && !trace_hash[:file].empty?
 
         # If we're going to send code then record the raw file path and the
--- a/lib/bugsnag/stacktrace.rb
+++ b/lib/bugsnag/stacktrace.rb
@@ -49,7 +49,6 @@ module Bugsnag
         end
 
         # Strip common gem path prefixes
-        # TODO test this
         if defined?(Gem)
           file = Gem.path.inject(file) {|line, path| line.sub(/#{path}\//, "") }
         end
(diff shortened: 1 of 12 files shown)


lib/bugsnag/stacktrace.rb
```diff
@@ -10,105 +11,63 @@ class Stacktrace
...
+          trace_hash.delete(:inProject) if file.match(configuration.vendor_path)
         end
 
-
         # Strip common gem path prefixes
         if defined?(Gem)
           file = Gem.path.inject(file) {|line, path| line.sub(/#{path}\//, "") }
         end
 
+        next if !file || file.empty?
```
kattrali, line 54: What case is this handling? The file path was equal to the project root / a gem path?

imjoehaines: Yeah exactly — [this test triggers it](https://github.com/bugsnag/bugsnag-ruby/blob/0fe026245b5e94939a3a407b856d8e1790dff56d/spec/stacktrace_spec.rb#L404-L425)

kattrali: Right but is that a thing that can happen? Backtrace lines that point to directories? If so, in the case that a line has a (non project root / gem path) directory, that should be something we think about in a few places.

imjoehaines: I don't think so, no. You could trigger it by setting the project root to a file rather than a directory too, which is a _little_ more realistic, but I don't think this is ever really triggered in practice

The only reason I kept it is because it [was here before](https://github.com/bugsnag/bugsnag-ruby/blob/59254960b973cbb8f635c29fe18526663b6c6a2c/lib/bugsnag/stacktrace.rb#L60-L64) and I can't say for certain that there's no reason for it. I went back through the history again and it was [implemented in this commit](https://github.com/bugsnag/bugsnag-ruby/commit/a479b6ed10ffdd3b808883037db92b27553aec6a) which also let users add their own filters for the stacktrace file name, so it would have been necessary then but seems like it's not anymore (unless the project root/a gem path point to a file)

kattrali: Hmmm ghosts in the code. I'd also suspect that its not needed anymore, especially since there are simpler mechanisms now for filtering out stacktrace lines if needed.

imjoehaines: spooky

👍 I'll remove it

PR #620, opened 2020-07-31
[PLAT-3702] Report runtime versions for Delayed Job, Mailman and Shoryuken

imjoehaines:
## Goal

This PR adds runtime versions for:

- Delayed Job
- Mailman
- Shoryuken

Mailman & Shoryuken are straightforward, but Delayed Job doesn't have a constant with its version in so instead we have to read it from the loaded Gems

## Review

  - [x] Final review

### DIFF
--- a/lib/bugsnag/integrations/delayed_job.rb
+++ b/lib/bugsnag/integrations/delayed_job.rb
@@ -10,11 +10,23 @@ end
 module Delayed
   module Plugins
     class Bugsnag < ::Delayed::Plugin
+      ##
+      # DelayedJob doesn't have an easy way to fetch its version, but we can use
+      # Gem.loaded_specs to get the version instead
+      def self.delayed_job_version
+        ::Gem.loaded_specs['delayed_job'].version.to_s
+      rescue StandardError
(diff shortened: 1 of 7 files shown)


lib/bugsnag/integrations/delayed_job.rb
```diff
@@ -10,11 +10,22 @@
 module Delayed
   module Plugins
     class Bugsnag < ::Delayed::Plugin
+      ##
+      # DelayedJob doesn't have an easy way to fetch its version, but we can use
+      # Gem.loaded_specs to get the version instead
+      def self.delayed_job_version
+        ::Gem.loaded_specs['delayed_job'].version.to_s
+      rescue StandardError
+        nil
```
imjoehaines, line 19: This is to make Rubocop happy — if we don't explicitly put a `nil` here it gets upset about suppressing exceptions

kattrali: This should get a little comment or something - I could easily see deleting the `nil` in the future because it doesn't normally matter.

PR #628, opened 2020-08-14
SUP-34  GitHub templates

steve-nester-uk:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,56 @@
...
+### Environment
+* Bugsnag version:
+* Ruby version:
+* Bundle version:
+* Integration framework version:
+    * Que:
+    * Rack
+    * Rails:
+    * Rake:
+    * Sidekiq:
```
imjoehaines, line 30: Are these just the integrations we have specific docs for? It might be worth adding the other integrations we have too, but they could also come under "other" (I'm happy either way, just raising the question!)

e.g.

```suggestion
    * DelayedJob
    * Mailman
    * Mongo
    * Que:
    * Rack
    * Rails:
    * Rake:
    * Resque:
    * Shoryuken:
    * Sidekiq:
```

steve-nester-uk: Yes, I took the list based on the docs https://docs.bugsnag.com/platforms/ruby/.  In this case personally I'd prefer to match the docs by implying the rest under 'others'.

PR #634, opened 2020-09-08
Workaround Bundler's friendly errors swallowing error reports

imjoehaines:
## Goal

When running a script with `bundle exec <file>` or `bundle exec ruby <file>`, Bundler will wrap any uncaught exceptions in a `SystemExit` — see https://github.com/rubygems/rubygems/issues/3368 and [`friendly_errors.rb`](https://github.com/rubygems/rubygems/blob/44388bfd7abad56f0f882fb9bc3f6c7c480f10bd/bundler/lib/bundler/friendly_errors.rb#L116-L123)

This means that the uncaught exceptions are ignored by default, as we don't report `SystemExit` exceptions, which can lead to real errors not being reported

This PR attempts to workaround this behaviour by trying to "unwrap" the original exception from the `SystemExit`. This is made more complicated by it being possible to have _two_ levels of `SystemExit` exceptions, if an executable script is run directly with `bundle exec <file>`. In this case we unwrap the "friendly errors" `SystemExit` and then unwrap the original exception

It's also important that we don't unwrap other `SystemExit` exceptions, because then we will report any time an exit is used in a `rescue` block, e.g.

```ruby
begin
  abc
  xyz
rescue
  exit
end
```

## Design

I'm not a huge fan of inspecting the backtrace for specific files, but they have stayed consistent for years so it should be fairly safe. I wasn't able to find another way to do this that didn't also result in reporting legitimate exits

## Changeset

- added `unwrap_bundler_exception` which is used in the `at_exit` hook
- reworked plain ruby maze runner tests

## Testing

The maze runner tests run the same files with the same results using both `bundle exec` and `bundle exec ruby`, which proves we can unwrap 1 or 2 levels of exceptions as needed

The new `exit_after_exception` test ensures we don't report exits outside of Bundler

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Enhancements
+
+* Bugsnag should now report uncaught exceptions inside Bundler's 'friendly errors'
+  | [#634](https://github.com/bugsnag/bugsnag-ruby/pull/634)
+
 ## 6.17.0 (27 August 2020)
 
 ### Enhancements
--- a/features/fixtures/plain/app/unhandled/custom_error.rb
+++ b/features/fixtures/plain/app/unhandled/custom_error.rb
@@ -1,7 +1,7 @@
-require './app'
+#!/usr/bin/env ruby
+require_relative '../app'
 
 configure_basics
-add_at_exit
 
 class CustomError < RuntimeError
 end
--- a/features/fixtures/plain/app/unhandled/exit_after_exception.rb
+++ b/features/fixtures/plain/app/unhandled/exit_after_exception.rb
@@ -1,0 +1,10 @@
+#!/usr/bin/env ruby
+require_relative '../app'
+
+configure_basics
+
+begin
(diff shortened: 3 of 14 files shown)


kattrali (approved):

deivid-rodriguez:
Hi! I'm fixing this in https://github.com/rubygems/rubygems/issues/3368.

Just letting you know in case you want to give the fix a try and see if it works for you, and also so that you are aware and can remove the workarounds you had to add.

Regards!

imjoehaines:
> Hi! I'm fixing this in [rubygems/rubygems#4063](https://github.com/rubygems/rubygems/pull/4063).
> 
> Just letting you know in case you want to give the fix a try and see if it works for you, and also so that you are aware and can remove the workarounds you had to add.
> 
> Regards!
> 
> EDIT: I linked to the original issue, not to the PR, updated now :)

Thanks for the heads up! I've done some testing today with Bundler 2.2.4 and your fix does make this workaround unnecessary

We'll keep it around for a while so people have a chance to update Bundler, but thanks for fixing the root issue 🙂

PR #636, opened 2020-10-08
Truncate Exception Messages if needed.

joshuapinter:
## Goal

If an exception occurs, such as a `Mysql2::Error::ConnectionError`, where the query is extremely long, this can cause the payload to exceed what Bugsnag will accept and throw a `Errno::EPIPE: Broken pipe` error when trying to send the request to Bugsnag.

A number of elements are already trimmed when the payload is too large, including meta data, stacktrace code, etc. However, the exception messages were not being trimmed.

This commit adds exception message trimming to `Bugsnag::Helpers.trim_if_needed`, which should help reduce errors when the queries causing the errors are exceptionally long.

## Design

A number of elements are already trimmed when the payload is too large, including meta data, stacktrace code, etc. However, the exception messages were not being trimmed.

This commit adds exception message trimming to `Bugsnag::Helpers.trim_if_needed`, which should help reduce errors when the queries causing the errors are exceptionally long.

## Changeset

Added `truncate_exception_messages` to `Bugsnag::Helpers.trim_if_needed` method, which truncates the exception messages if necessary.

## Testing

Manually on our Rails 4.2 / Bugsnag 6.17.0 enterprise application. 

Added a spec to `helper_spec.rb` that tests this functionality specifically.

✅  All tests passing when running `bundle exec rake`.

### DIFF
--- a/lib/bugsnag/helpers.rb
+++ b/lib/bugsnag/helpers.rb
@@ -19,8 +19,12 @@ module Bugsnag
 
       return value unless payload_too_long?(value)
 
+      # Truncate exception messages
+      reduced_value = truncate_exception_messages(value)
+      return reduced_value unless payload_too_long?(reduced_value)
+
       # Trim metadata
-      reduced_value = trim_metadata(value)
+      reduced_value = trim_metadata(reduced_value)
       return reduced_value unless payload_too_long?(reduced_value)
 
       # Trim code from stacktrace
@@ -72,6 +76,15 @@ module Bugsnag
     TRUNCATION_INFO = '[TRUNCATED]'
 
     ##
+    # Truncate exception messages
+    def self.truncate_exception_messages(payload)
+      extract_exception(payload) do |exception|
+        exception[:message] = trim_as_string(exception[:message])
+      end
+      payload
+    end
+
+    ##
     # Remove all code from stacktraces
     def self.trim_stacktrace_code(payload)
       extract_exception(payload) do |exception|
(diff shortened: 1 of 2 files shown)


imjoehaines:
Hey @joshuapinter, thanks for the PR

This one will need a bit of discussion internally as we aim to have a single algorithm that all of our libraries use to ensure a payload is always less than the maximum length (for historical reasons, bugsnag-ruby doesn't quite implement that yet!)

I think this change makes sense though — it doesn't seem like there would ever be much value in a >3000 character exception message — but we'll need to make sure we've thought through everything before merging, as well as the nuances like what order we should trim things in

In the meantime, you can implement this using an `add_on_error` callback:

```ruby
Bugsnag.configuration.add_on_error(proc do |report|
    next unless exception[:message].length > 3000

    exception[:message] = exception[:message][0...3000]
  end
end)
```

joshuapinter:
Sounds good @imjoehaines. For now we're just using our fork of it. Whether this change gets merged or if the standardized trimming algorithm includes it, it's an important change. We had multiple customers running into serious issues without knowing it and it was only the couple customers that submitted a Support Request that made us aware of it. 

Let me know if you need anything from our end. 

Thanks!

imjoehaines:
Hey @joshuapinter, I've fixed the conflict and will merge this in https://github.com/bugsnag/bugsnag-ruby/pull/646

Thanks again for the PR — this change turned out to be super important for Ruby 3 as `NameError` and `NoMethodError` messages don't get truncated by default anymore (https://github.com/ruby/ruby/pull/3090)

joshuapinter:
Good timing then! Thanks for the update and have a great holiday and New Year.

PR #637, opened 2020-10-11
Add Note for installing test group gems.

joshuapinter:
Added a note that in order to run the specs, you may need to install the bundle with the `--with test` command line option in order to install the `test` group of gems, including `rake`. Otherwise you end up with an error like this: 

```bash
$ be rake spec

bundler: failed to load command: rake (/Users/joshuapinter/.rbenv/versions/2.6.5/bin/rake)
Gem::Exception: can't find executable rake for gem rake. rake is not currently included in the bundle, perhaps you meant to add it to your Gemfile?
```

Wasted 15 minutes figuring out what the issue was so this should help others contribute.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,9 @@ Thank you!
     ```
     bundle exec rake spec
     ```
+    
+    _NOTE: You may need to install the bundle with the `test` group first by running `bundle install --with test`. Otherwise, you will get `can't find executable rake for gem rake` error._
+    
 * Further information on installing and running the tests can be found in [the testing guide](TESTING.md)
 
 ### Document
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,12 +15,15 @@ Thank you!
 
 ### Polish
 
+* Install the test dependencies
+    ```
+    bundle install --with test sidekiq --binstubs
+    ```
+
 * Run the tests with and make sure they all pass
     ```
     bundle exec rake spec
     ```
-    
-    _NOTE: You may need to install the bundle with the `test` group first by running `bundle install --with test`. Otherwise, you will get `can't find executable rake for gem rake` error._
     
 * Further information on installing and running the tests can be found in [the testing guide](TESTING.md)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ Thank you!
 
 * Run the tests with and make sure they all pass
     ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -19,6 +19,9 @@ Thank you!
     ```
     bundle exec rake spec
     ```
+    
+    _NOTE: You may need to install the bundle with the `test` group first by running `bundle install --with test`. Otherwise, you will get `can't find executable rake for gem rake` error._
```
imjoehaines, line 23: I think it makes sense to have the installation step as its own bullet point in the contributing guide as hopefully then contributors won't see the error at all. So maybe something like this as the first bullet in this "Polish" section would be better:

```markdown
* Install the test dependencies
    ```
    bundle install --with test sidekiq --binstubs
    ```
```

imjoehaines (changes requested):
Hey @joshuapinter, thanks for the PR!

There is a bundle install command in [TESTING.md](https://github.com/bugsnag/bugsnag-ruby/blob/master/TESTING.md) already, but it's not clear enough that it's a required setup step (it looks like part of the command to run the tests instead), so I think this change makes sense

As it's required before the command to run the tests will work, I think it makes sense to be its own bullet point, especially as there shouldn't be any harm in re-running it

I'll have a look at updating the testing guide to be clearer about the setup as well

CONTRIBUTING.md
joshuapinter: Also, in the README it says the command to run the tests is `bundle exec rake spec` whereas the TESTING file says `bundle exec rake`. Which one do you prefer?

joshuapinter:
K, updated to a separate bullet point for installing test bundle.

CONTRIBUTING.md
imjoehaines: I think following the testing guide makes sense, so `bundle exec rake` — `spec` is the default so there's no real difference 🙂

TESTING.md isn't mentioned in the readme, but I think that's OK after this PR. If all the required steps for the unit tests are in CONTRIBUTING (which is mentioned in the readme) then the rest of the testing guide is irrelevant for outside contributors as the end to end tests can only be run with private credentials

joshuapinter:
Also changed `bundle exec rake spec` to just `bundle exec rake` for consistency with `TESTING.md`.

imjoehaines (approved):
Thanks @joshuapinter!

johnkiely1:
This has now been added in v6.18.0.

joshuapinter:
👍

PR #638, opened 2020-10-20
Fix CI issues

imjoehaines:
Our CI pipeline is currently broken due to a double whammy of Rubocop and Maze Runner bugs 😀

Rubocop 0.88.0 has an issue with the LineLength cop that causes it to error; I'm not sure why we weren't seeing this before, but this is fixed by moving to the next version (0.89). There are a couple of related changes — new cops are added to rubocop.yml and there's a small change in `helpers.rb` to fix an issue for one of the new cops

Maze Runner 2.6.0 added a hook to clear the environment between test runs (see https://github.com/bugsnag/maze-runner/pull/130). This breaks our CI because it happens _after_ we set the API key and endpoint, so none of the tests can report anything. MR v3.0.2 fixes this, so we need to update

Finally, some of the test dockerfiles were using the wrong argument and so didn't work correctly. Previously these ran the tests and passed (we don't really know why) but failed on MR v3. The dockerfiles have been fixed and they now run as expected. This found an issue with Mailman & Que on Ruby 2.0, which will be fixed separately (they've been skipped for now)


.buildkite/pipeline.yml
```diff
@@ -585,6 +585,7 @@ steps:
       RUBY_TEST_VERSION: '2.0'
     concurrency: 4
     concurrency_group: 'ruby/integrations-maze-runner-tests'
+    if: "false"
```
twometresteve, line 588: It's the same effect and I'm not overly bothered, but Buildkite has a `skip` that may be better - for example if we wanted to audit skipped tests across the pipelines it would be easier to find.  This is one from Android:
```    
skip: Temporarily disabled due to flakiness with BrowserStack tunneling
```

You may also want to consider a "soft fail", where the test is still run but doesn't cause the pipeline to have an overall fail.  It's wasted energy, perhaps, but the result is more in your face.  I guess it depends how much you trust yourself to work fix the failures in time.
```
    soft_fail:
      - exit_status: "*"
```

Gemfile
```diff
@@ -32,7 +32,7 @@ group :coverage, optional: true do
 end
 
 group :rubocop, optional: true do
-  gem 'rubocop', '~> 0.88.0'
+  gem 'rubocop', '~> 0.89.0'
```
twometresteve, line 35: I was going to ask why not `0.93.1`, but in fact `1.0.0` has just been released today.

dockerfiles/Dockerfile.ruby-maze-runner
```diff
@@ -20,7 +20,7 @@ COPY spec ./spec
 RUN gem build bugsnag.gemspec
 
 # The maze-runner node tests
-FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:latest-cli as ruby-maze-runner
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:v3.0.2-cli as ruby-maze-runner
```
twometresteve, line 23: Yeah, fine to pin to a version, but I would request that you keep an eye out for new releases and consider upgrading occasionally as not to be left too far behind.

twometresteve (approved):
I've left a couple of suggestions, but happy for you to make the call on implementing them.

Gemfile
imjoehaines: I find Rubocop updating is quite a lot of work for not much gain, so I didn't want to jump more versions than required to get CI going again. I'll bump to 1.0 though, as it should now be a lot more stable

.buildkite/pipeline.yml
imjoehaines: Ah good to know; I didn't see `skip` in the docs. Soft fail sounds good, I'll do that instead

Gemfile
twometresteve: Fingers crossed!  Of course I'd be happy to leave it at 0.89 and factor out the move to 1.0 for another day if it's not straightforward.

imjoehaines: Turns out to have been super easy & it caught a duplicate `require` 😀

dockerfiles/Dockerfile.ruby-maze-runner
imjoehaines: Yeah, I just want to avoid the same thing happening again; this was quite painful to debug, especially with Rubocop failing at the same time :s

twometresteve: 👍  Once bitten...

twometresteve (approved):
LGTM pending CI.

PR #646, opened 2020-12-21
Truncate Exception Messages if needed

imjoehaines:
This is PR #636 rebased onto `next` + a changelog entry

---

## Goal

If an exception occurs, such as a `Mysql2::Error::ConnectionError`, where the query is extremely long, this can cause the payload to exceed what Bugsnag will accept and throw a `Errno::EPIPE: Broken pipe` error when trying to send the request to Bugsnag.

A number of elements are already trimmed when the payload is too large, including meta data, stacktrace code, etc. However, the exception messages were not being trimmed.

This commit adds exception message trimming to `Bugsnag::Helpers.trim_if_needed`, which should help reduce errors when the queries causing the errors are exceptionally long.

## Design

A number of elements are already trimmed when the payload is too large, including meta data, stacktrace code, etc. However, the exception messages were not being trimmed.

This commit adds exception message trimming to `Bugsnag::Helpers.trim_if_needed`, which should help reduce errors when the queries causing the errors are exceptionally long.

## Changeset

Added `truncate_exception_messages` to `Bugsnag::Helpers.trim_if_needed` method, which truncates the exception messages if necessary.

## Testing

Manually on our Rails 4.2 / Bugsnag 6.17.0 enterprise application. 

Added a spec to `helper_spec.rb` that tests this functionality specifically.

✅  All tests passing when running `bundle exec rake`.

### DIFF
--- a/spec/helper_spec.rb
+++ b/spec/helper_spec.rb
@@ -41,9 +41,26 @@ describe Bugsnag::Helpers do
     end
 
     context "payload length is greater than allowed" do
+      it "trims exception messages" do
+        payload = {
+          :events => [{
+            :exceptions => [{
+              :message => 50000.times.map {|i| "should truncate" }.join(""),
+              :preserved => "Foo"
+            }]
+          }]
+        }
+        expect(::JSON.dump(payload).length).to be > Bugsnag::Helpers::MAX_PAYLOAD_LENGTH
+        trimmed = Bugsnag::Helpers.trim_if_needed(payload)
+        expect(::JSON.dump(trimmed).length).to be <= Bugsnag::Helpers::MAX_PAYLOAD_LENGTH
+        expect(trimmed[:events][0][:exceptions][0][:message].length).to be <= Bugsnag::Helpers::MAX_STRING_LENGTH
+        expect(trimmed[:events][0][:exceptions][0][:preserved]).to eq("Foo")
+      end
+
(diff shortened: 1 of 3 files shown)


spec/helper_spec.rb
```diff
@@ -41,9 +41,26 @@
...
+            :exceptions => [{
+              :message => 50000.times.map {|i| "should truncate" }.join(""),
+              :preserved => "Foo"
+            }]
+          }]
+        }
+        expect(::JSON.dump(payload).length).to be > Bugsnag::Helpers::MAX_PAYLOAD_LENGTH
+        trimmed = Bugsnag::Helpers.trim_if_needed(payload)
+        expect(::JSON.dump(trimmed).length).to be <= Bugsnag::Helpers::MAX_PAYLOAD_LENGTH
+        expect(trimmed[:events][0][:exceptions][0][:message].length).to be <= Bugsnag::Helpers::MAX_STRING_LENGTH
```
kattrali, line 56: This test doesn't verify what the value is, in case its been converted to an empty string or shortened more than needed or something.

imjoehaines: Improved the assertions in https://github.com/bugsnag/bugsnag-ruby/pull/646/commits/0a2101724193d17af7b120d75d939f7295d1c90a

kattrali (approved):

PR #647, opened 2020-12-21
Add Ruby 3.0 to CI

imjoehaines:
**Waiting for full release & a 3.0.0 Docker tag before merging**

## Goal

Adds Ruby 3.0 to CI, with related changes to fixtures to make this work

## Changeset

- Added Gems that are no longer in the stdlib to the `test` group in `Gemfile`
- Ensure we get a version of 'redis-namespace' that's compatible with Ruby 3 for the Rails integrations tests (this is temporary until a new version is released)
- Fix an issue with the "I navigate to the route {string} on the rails app" step in Ruby 3
- Use minitest 5.14 in fixture apps
- Fix ruby version comparison helper

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -194,6 +194,17 @@ steps:
     concurrency: 4
     concurrency_group: 'ruby/unit-tests'
 
+  - label: ':ruby: Ruby 3.0 unit tests'
+    timeout_in_minutes: 30
+    plugins:
+      docker-compose#v3.1.0:
+        run: ruby-unit-tests
+        use-aliases: true
+    env:
+      RUBY_TEST_VERSION: "3.0"
+    concurrency: 4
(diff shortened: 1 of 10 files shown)


.buildkite/pipeline.yml
```diff
@@ -194,6 +194,17 @@ steps:
...
     concurrency_group: 'ruby/unit-tests'
 
+  - label: ':ruby: Ruby 3.0 unit tests'
+    timeout_in_minutes: 30
+    plugins:
+      docker-compose#v3.1.0:
+        run: ruby-unit-tests
+        use-aliases: true
+    env:
+      RUBY_TEST_VERSION: "3.0.0-preview2"
```
imjoehaines, line 204: TODO: this version needs changing to 3.0 when it's released & there's a tag on docker hub

kattrali (approved):

PR #654, opened 2021-03-23
Classify `ActionDispatch::Http::MimeNegotiation::InvalidType` errors as `info`-level severity

Laukei:
## Goal

[Rails 6.1 adds a specific error](https://github.com/rails/rails/pull/40353) for [`Mime::Type::InvalidMimeType`s](https://github.com/rails/rails/issues/37620) primarily for error trackers, as the error is handled by Rails and returns a HTTP 406 response. This PR catches instances of that error as info-level severity, in keeping with existing behaviour for other similar errors.

## Changeset

- The exception was added to the `INFO_CLASSES` in `Bugsnag::Middleware::ClassifyError`

## Testing

- None


imjoehaines (approved):
LGTM

Can you add a changelog entry for this too?

Laukei:
> Can you add a changelog entry for this too?

Done 👍

PR #655, opened 2021-03-26
Include the connection ID in ActiveRecord breadcrumbs on new version of Rails

imjoehaines:
## Goal

Rails replaced the connection ID with a connection object in ActiveRecord instrumentation. We will now get the ID from the connection object, as before

This is partly tested by existing tests and is tested against Rails 6.1 in https://github.com/bugsnag/bugsnag-ruby/pull/656

### DIFF
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -91,10 +91,21 @@ module Bugsnag
         filtered_data = data.slice(*event[:allowed_data])
         filtered_data[:event_name] = event[:id]
         filtered_data[:event_id] = event_id
-        if event[:id] == "sql.active_record" && data.key?(:binds)
-          binds = data[:binds].each_with_object({}) { |bind, output| output[bind.name] = '?' if defined?(bind.name) }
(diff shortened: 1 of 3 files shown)


lib/bugsnag/integrations/railtie.rb
```diff
@@ -91,10 +91,19 @@ def event_subscription(event)
...
-          binds = data[:binds].each_with_object({}) { |bind, output| output[bind.name] = '?' if defined?(bind.name) }
-          filtered_data[:binds] = JSON.dump(binds) unless binds.empty?
+
+        if event[:id] == "sql.active_record"
+          if data.key?(:binds)
+            binds = data[:binds].each_with_object({}) { |bind, output| output[bind.name] = '?' if defined?(bind.name) }
+            filtered_data[:binds] = JSON.dump(binds) unless binds.empty?
+          end
+
+          if data.key?(:connection) && !data.key?(:connection_id)
```
kattrali, line 101: Might be worth annotating `connection_id` here or in the `allowed_data` list to show its for older versions.

kattrali (approved):

PR #658, opened 2021-04-14
Add license checker to CI

twometresteve:
## Goal

Adds license checking to CI, following the same approach of our other Ruby based repos.

## Testing

Covered by CI.  Requires https://github.com/bugsnag/license-audit/pull/8 to be merged for the new step to pass.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,12 +11,6 @@ steps:
             - ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby${BRANCH_NAME}
             - ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby-latest
 
-  - name: ':copyright: License Audit'
-    plugins:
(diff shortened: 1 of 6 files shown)


.buildkite/pipeline.yml
```diff
@@ -11,6 +11,12 @@ steps:
             - ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby${BRANCH_NAME}
             - ruby-maze-runner:855461928731.dkr.ecr.us-west-1.amazonaws.com/ruby:base-ruby-latest
 
+  - name: ':copyright: License Audit'
+    plugins:
+      docker-compose#v3.7.0:
+        run: license_finder
+    command: /bin/bash -lc '/scan/scripts/license_finder.sh'
```
imjoehaines, line 18: Can we either put this in the first or second group of test steps rather than here? Otherwise we won't get feedback from CI if we introduce a dependency that doesn't pass the audit, until it's approved

twometresteve: Yes, that makes sense.

imjoehaines (approved):

PR #660, opened 2021-06-02
Deliver even when an error raised in the block argument

aki77:
## Goal

I would also like to be notified if there is an error in the block argument of the notify method.
This is because the more important the error handling, the more complex the contents of the block argument, and the more likely it is that an error will occur.

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -106,7 +106,11 @@ module Bugsnag
 
         # If this is not an auto_notify then the block was provided by the user. This should be the last
         # block that is run as it is the users "most specific" block.
-        yield(report) if block_given? && !auto_notify
+        begin
+          yield(report) if block_given? && !auto_notify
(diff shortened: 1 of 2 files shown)


johnkiely1:
Hey @aki77, 
Thanks for the PR, we will look to review as soon as priorities allow.

imjoehaines:
Thanks for the PR @aki77, this is a really useful fix!

I'll merge this in https://github.com/bugsnag/bugsnag-ruby/pull/664 where I've made a change to log a warning instead of adding metadata. This is what we do in other places where we rescue errors from user-supplied functions, e.g. in `on_error` callbacks:

https://github.com/bugsnag/bugsnag-ruby/blob/98b73dc7632ac99055abd3d3a2af1f7b901b4ee5/lib/bugsnag/on_error_callbacks.rb#L13-L18

aki77:
Thanks!

PR #670, opened 2021-08-03
Add support for Active Job 

imjoehaines:
## Goal

This PR adds support for Active Job when not using an existing integration. For example, nothing will change when using Sidekiq as the backend but we can now track exceptions in Active Job's `async` backend

We capture the following metadata automatically for both handled and unhandled exceptions (note some data is not available on certain Rails versions):

- `job_id`
- `job_name`
- `queue`
- `locale`
- `arguments`
- `provider_job_id` (added in Rails 5)
- `executions` (added in Rails 5)
- `timezone` (added in Rails 6)
- `enqueued_at` (added in Rails 6)

The context is set to the job name and queue to match other queue integrations, e.g. `ExampleJob@some_queue`, and the default app type is set to "active job"

## Testing

Maze Runner tests run against Rails 4, 5 & 6 with a variety of Ruby versions

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,14 @@ Lint/RescueException:
   Exclude:
     - 'lib/bugsnag/integrations/**/*'
 
+Metrics/AbcSize:
+  Exclude:
+    - 'lib/bugsnag/configuration.rb'
+
+Metrics/ClassLength:
+  Exclude:
+    - 'lib/bugsnag/configuration.rb'
+
 # We can't use ".freeze" on our constants in case users are monkey patching
 # them — this would be a BC break
 Style/MutableConstant:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,10 @@ AllCops:
     - 'spec/fixtures/**/*'
     - 'features/fixtures/**/*'
 
+Lint/RescueException:
+  Exclude:
+    - 'lib/bugsnag/integrations/**/*'
+
 # We can't use ".freeze" on our constants in case users are monkey patching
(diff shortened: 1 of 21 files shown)


kattrali (approved):

knu:
Hi,

After this change, exceptions handled by `discard_on` seem to get notified.  Is that intentional?  I'd expect discarded exceptions not to be reported.

imjoehaines:
> After this change, exceptions handled by `discard_on` seem to get notified. Is that intentional? I'd expect discarded exceptions not to be reported.

Yes, this is expected as there isn't an API to detect if an error would be discarded. An error being discarded also doesn't necessarily mean it's unimportant to know about, since errors that would make a job impossible to finish will often end up being discarded to prevent retries that will never succeed

If you have errors that you don't want to end up in Bugsnag, you can use the [`discard_classes`](https://docs.bugsnag.com/platforms/ruby/rails/configuration-options/#discard_classes) option to discard an error by name. If the logic needs to be more complicated then you can use [an on error callback](https://docs.bugsnag.com/platforms/ruby/rails/configuration-options/#add_on_error) instead. For example, to ignore a specific combination of error and job:

```ruby
Bugsnag.configure do |config|
  config.add_on_error(proc do |report|
    next if report.exceptions.empty?
    next if report.meta_data[:active_job].nil?

    error_class = report.exceptions.first[:errorClass]
    job_name = report.meta_data[:active_job][:job_name]

    if error_class == 'ClassToIgnore' && job_name == 'JobToIgnore'
      false # ignore the report
    end
  end)
end
```

PR #676, opened 2021-08-12
Deprecate and provide replacements for inconsistent methods

imjoehaines:
## Goal

For consistency with Bugsnag notifiers for other languages, a number of methods have been deprecated in this PR. The old options will be removed in the next major version

* The `notify_release_stages` configuration option has been deprecated in favour of `enabled_release_stages`
* The `auto_capture_sessions` and `track_sessions` configuration options have been deprecated in favour of `auto_track_sessions`
* The `Report` class has been deprecated in favour of the `Event` class
* The `Report#meta_data` attribute has been deprecated in favour of `Report#metadata`
* The `Breadcrumb#meta_data` attribute has been deprecated in favour of `Breadcrumb#metadata`
* The `Breadcrumb#name` attribute has been deprecated in favour of `Breadcrumb#message`
* The breadcrumb type constants in the `Bugsnag::Breadcrumbs` module has been deprecated in favour of the constants available in the `Bugsnag::BreadcrumbType` module
For example, `Bugsnag::Breadcrumbs::ERROR_BREADCRUMB_TYPE` is now available as `Bugsnag::BreadcrumbType::ERROR`

## Design

I've avoided using the `alias` keyword as YARD will merge the alias into the original attribute, e.g the "also" here:

|image|
|-|

This makes it impossible to mark the original name as deprecated without also marking the new name as deprecated<sup>1</sup>:

|image|
|-|

Therefore I've implemented the aliases manually to give them their own section in the generated API docs:

|image|
|-|

The new methods use [the `@!attribute` directive](https://rubydoc.info/gems/yard/file/docs/Tags.md#attribute) so YARD generates the docs as if they were attributes and not methods

<sup>1</sup> I believe you _can_ do it as it's a behaviour of the templates YARD uses, but usually aliases being merged is correct so it doesn't seem worth writing new templates just for this PR


CHANGELOG.md
```diff
@@ -1,6 +1,20 @@
...
 
+## TBD
+
+### Deprecated
+
+* For consistency with Bugsnag notifiers for other languages, a number of methods have been deprecated in this release. The old options will be removed in the next major version | [#676](https://github.com/bugsnag/bugsnag-ruby/pull/676)
+    * The `notify_release_stages` configuration option has been deprecated in favour of `enabled_release_stages`
+    * The `auto_capture_sessions` and `track_sessions` configuration options have been deprecated in favour of `auto_track_sessions`
+    * The `Report` class has been deprecated in favour of the `Event` class
+    * The `Report#meta_data` attribute has been deprecated in favour of `Report#metadata`
```
kattrali, line 12:
```suggestion
    * The `Report#meta_data` attribute has been deprecated in favour of `Event#metadata` ?
```

kattrali (approved):

PR #677, opened 2021-08-16
Don't start sessions when the current release stage is not enabled

imjoehaines:
## Goal

Currently we start sessions but will not deliver them when the current release stage is not enabled

This PR changes the session tracker to avoid starting a session instead; this has a small performance benefit as the background delivery thread won't be started for sessions that will never be delivered


kattrali (approved):
nice one!

PR #681, opened 2021-08-18
Log delivery failures as errors

imjoehaines:
## Goal

Currently we log exceptions in delivery as warnings, which is the same level that we log recoverable exceptions (e.g. exceptions in `on_error` callbacks). This makes it easy to miss log messages that mean delivery has totally failed as they are mixed in with other messages of the same level

This PR switches delivery failures to log at the error level so that they are more noticeable and to allow recoverable log messages to be suppressed without also suppressing unrecoverable errors

## Testing

I've refactored the current logging tests to use a real `Logger` object, which is the first commit. [Use this comparison](https://github.com/bugsnag/bugsnag-ruby/compare/9bf9dd80229b72986f6654786afea10f97a4057f...693a6662b4a294ce9013641938a2985fde2481f6) for only the delivery log level changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Changelog
 
 * Sessions will now be delivered every 10 seconds, instead of every 30 seconds
   | [#680](https://github.com/bugsnag/bugsnag-ruby/pull/680)
+* Log errors that prevent delivery at `ERROR` level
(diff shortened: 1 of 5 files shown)


kattrali (approved):
The switch to an actual `Logger` is a really nice improvement :+1:

PR #688, opened 2021-08-25
Give the configuration context precedence over context in integrations

imjoehaines:
## Goal

If `Configuration#context` is set then it will now take precedence over the context set by integrations. This is true even if it is set to `nil`, in which case no context will be sent by the notifier — the context will be set by the Bugsnag app instead

For example:

image

The top error has `config.context = nil`, so the context is set by the Bugsnag app
The middle error has `config.context = 'some context'` and so always uses that value as the context
The bottom error has no context set in `config`, which allows the Rack integration's context to be used


kattrali (approved):
This is a nice solution to a hairy problem!

PR #701, opened 2021-09-30
Bring endpoint configuration in line with other notifiers

imjoehaines:
## Goal

This PR introduces `Configuration#endpoints`/`Configuration#endpoints=`, bringing endpoint configuration in line with other notifiers (e.g. [Cocoa](https://docs.bugsnag.com/platforms/ios/configuration-options/#endpoints), [Android](https://docs.bugsnag.com/platforms/android/configuration-options/#endpoints) & [JS](https://docs.bugsnag.com/platforms/javascript/configuration-options/#endpoints))

Setting the endpoints is done with a new `Bugsnag::EndpointConfiguration` class:

```ruby
Bugsnag.configure do |config|
  config.endpoints = Bugsnag::EndpointConfiguration.new(
    "<notify endpoint>",
    "<sessions endpoint>"
  )
end
```

The endpoints can then be read from `Configuration#endpoints`:

```ruby
puts Bugsnag.configuration.endpoints.notify # => <notify endpoint>
puts Bugsnag.configuration.endpoints.sessions # => <sessions endpoint>
```

When the endpoints are set, we validate that two non-empty strings have been given (i.e. `nil` & `""` fail validation, all other values pass). If validation fails then we will avoid sending requests to prevent leaking data to the cloud Bugsnag instance. The exception to this rule is that we will still send events if the notify endpoint is set without the sessions endpoint. This is for backwards compatibility — it's entirely possible for users of Bugsnag before sessions existed to still only be setting the notify endpoint. This exception will be removed in the next major version; all requests will be avoided if either URL is not valid

### DIFF
--- a/lib/bugsnag/endpoint_validator.rb
+++ b/lib/bugsnag/endpoint_validator.rb
@@ -22,12 +22,12 @@ module Bugsnag
     class Result
       # rubocop:disable Layout/LineLength
       MISSING_URLS = "Invalid configuration. endpoints must be set with both a notify and session URL. Bugsnag will not send any requests.".freeze
-      MISSING_NOTIFY_URL = "Invalid configuration. endpoints.notify cannot be set without also setting endpoints.sessions. Sessions will not be sent to Bugsnag.".freeze
-      MISSING_SESSION_URL = "Invalid configuration. endpoints.sessions cannot be set without also setting endpoints.notify. Bugsnag will not send any requests.".freeze
+      MISSING_NOTIFY_URL = "Invalid configuration. endpoints.sessions cannot be set without also setting endpoints.notify. Bugsnag will not send any requests.".freeze
+      MISSING_SESSION_URL = "Invalid configuration. endpoints.notify cannot be set without also setting endpoints.sessions. Bugsnag will not send any sessions.".freeze
 
       INVALID_URLS = "Invalid configuration. endpoints should be valid URLs, got empty strings. Bugsnag will not send any requests.".freeze
-      INVALID_NOTIFY_URL = "Invalid configuration. endpoints.notify should be a valid URL, got empty string. Sessions will not be sent to Bugsnag.".freeze
-      INVALID_SESSION_URL = "Invalid configuration. endpoints.sessions should be a valid URL, got empty string. Bugsnag will not send any requests.".freeze
+      INVALID_NOTIFY_URL = "Invalid configuration. endpoints.notify should be a valid URL, got empty string. Bugsnag will not send any requests.".freeze
+      INVALID_SESSION_URL = "Invalid configuration. endpoints.sessions should be a valid URL, got empty string. Bugsnag will not send any sessions.".freeze
       # rubocop:enable Layout/LineLength
 
       attr_reader :reason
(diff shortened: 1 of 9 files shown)


lib/bugsnag/endpoint_validator.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      Result.valid
+    end
+
+    # @api private
+    class Result
+      # rubocop:disable Layout/LineLength
+      MISSING_URLS = "Invalid configuration. endpoints must be set with both a notify and session URL. Bugsnag will not send any requests.".freeze
+      MISSING_NOTIFY_URL = "Invalid configuration. endpoints.notify cannot be set without also setting endpoints.sessions. Sessions will not be sent to Bugsnag.".freeze
+      MISSING_SESSION_URL = "Invalid configuration. endpoints.sessions cannot be set without also setting endpoints.notify. Bugsnag will not send any requests.".freeze
```
tomlongridge, line 26: I think the wording for these is the wrong way around? e.g. if the notify URL is missing, the message should say:

> endpoints.sessions cannot be set without also setting endpoints.notify.

lib/bugsnag/endpoint_validator.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    # @api private
+    class Result
+      # rubocop:disable Layout/LineLength
+      MISSING_URLS = "Invalid configuration. endpoints must be set with both a notify and session URL. Bugsnag will not send any requests.".freeze
+      MISSING_NOTIFY_URL = "Invalid configuration. endpoints.sessions cannot be set without also setting endpoints.notify. Bugsnag will not send any requests.".freeze
+      MISSING_SESSION_URL = "Invalid configuration. endpoints.notify cannot be set without also setting endpoints.sessions. Sessions will not be sent to Bugsnag.".freeze
+
+      INVALID_URLS = "Invalid configuration. endpoints should be valid URLs, got empty strings. Bugsnag will not send any requests.".freeze
+      INVALID_NOTIFY_URL = "Invalid configuration. endpoints.notify should be a valid URL, got empty string. Sessions will not be sent to Bugsnag.".freeze
```
tomlongridge, line 29:
```suggestion
      INVALID_NOTIFY_URL = "Invalid configuration. endpoints.notify should be a valid URL, got empty string. Bugsnag will not send any requests.".freeze
```

lib/bugsnag/endpoint_validator.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      Result.valid
+    end
+
+    # @api private
+    class Result
+      # rubocop:disable Layout/LineLength
+      MISSING_URLS = "Invalid configuration. endpoints must be set with both a notify and session URL. Bugsnag will not send any requests.".freeze
+      MISSING_NOTIFY_URL = "Invalid configuration. endpoints.sessions cannot be set without also setting endpoints.notify. Bugsnag will not send any requests.".freeze
+      MISSING_SESSION_URL = "Invalid configuration. endpoints.notify cannot be set without also setting endpoints.sessions. Sessions will not be sent to Bugsnag.".freeze
```
tomlongridge, line 26: I think this mixes the target of who "Bugsnag" is - i.e. the notifier or the endpoint - compared to the other messages.

```suggestion
      MISSING_SESSION_URL = "Invalid configuration. endpoints.notify cannot be set without also setting endpoints.sessions. Bugsnag will not send any sessions.".freeze
```

tomlongridge (approved):

PR #703, opened 2021-10-04
Add support for "redacted keys"

imjoehaines:
## Goal

This PR adds a new `redacted_keys` configuration option, which will replace `meta_data_filters` in the next major version

As well as being consistent with other notifier APIs, `redacted_keys` uses a case-insensitive, exact comparison. For example, the string `abc` will match `abc` and `ABC` but not `abcd`

This makes string matching more useful as it's less likely to hit a false positive, but the old behaviour is still possible by using a `Regexp`


spec/cleaner_spec.rb
```diff
@@ -276,6 +276,134 @@ def id
...
+      end
+
+      it "redacts deeply nested keys" do
+        configuration = Bugsnag::Configuration.new
+        configuration.redacted_keys = Set[/^foo\.bar/]
+
+        cleaner = Bugsnag::Cleaner.new(configuration)
+
+        params = { events: { metaData: { foo: { bar: 'baz' } } } }
+        expect(cleaner.clean_object(params)).to eq({ events: { metaData: { foo: { bar: '[FILTERED]' } } } })
```
tomlongridge, line 335: Would it be worth a negative check here too - e.g. add a second field and check that it's not filtered

spec/cleaner_spec.rb
```diff
@@ -276,6 +276,134 @@ def id
...
 
+    describe "redacted keys" do
+      it "redacts exact string match" do
+        object = { events: { metaData: { foo: 'bar' } } }
+
+        configuration = Bugsnag::Configuration.new
+        configuration.redacted_keys = Set['foo']
+
+        cleaner = Bugsnag::Cleaner.new(configuration)
+        expect(cleaner.clean_object(object)).to eq({ events: { metaData: { foo: '[FILTERED]' } } })
```
tomlongridge, line 287: This is probably going to complicate the change a lot, but the notifier spec string is "[REDACTED]" (not "FILTERED").

lib/bugsnag/configuration.rb
```diff
@@ -201,6 +207,7 @@ def initialize
       self.send_environment = false
       self.send_code = true
       self.meta_data_filters = Set.new(DEFAULT_META_DATA_FILTERS)
+      @redacted_keys = Set.new
```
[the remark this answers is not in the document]
tomlongridge: ... in fact maybe that is correct before `meta_data_filters` is removed to avoid a change in behaviour?

Then again, perhaps the longer list of defaults is correct for Ruby/Rails?

spec/report_spec.rb
```diff
@@ -1155,6 +1155,62 @@ def gloops
...
+      })
+    end
+
+    expect(Bugsnag).to have_sent_notification{ |payload, headers|
+      event = get_event_from_payload(payload)
+      expect(event["metaData"]).not_to be_nil
+      expect(event["metaData"]["request"]).not_to be_nil
+      expect(event["metaData"]["request"]["params"]).not_to be_nil
+      expect(event["metaData"]["request"]["params"]["password"]).to eq("[FILTERED]")
+      expect(event["metaData"]["request"]["params"]["other_password"]).to eq("[FILTERED]")
```
tomlongridge, line 1179: Shouldn't the match for string keys be exact, so not match "other_password" or "more_other_data"?

(Also, should we assert that in the cleaner spec?)

lib/bugsnag/configuration.rb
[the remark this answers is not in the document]
imjoehaines: We can't add defaults for redacted keys until `meta_data_filters` is removed, otherwise you'd have to remove the value from both to prevent it being redacted/filtered

spec/cleaner_spec.rb
imjoehaines: Yeah I figured that wasn't particularly important — it can changed to say "[REDACTED]" when `meta_data_filters` are removed

spec/report_spec.rb
imjoehaines: It is exact — `other_password` & `password` are being filtered by the default `meta_data_filters`

`other_data` is filtered by `redacted_keys` and `more_other_data` is not, which proves it's exact

tomlongridge: Ah - got it now. OK, I guess that simplifies significantly in major release.

lib/bugsnag/configuration.rb
```diff
@@ -56,9 +56,15 @@ class Configuration
 
     # A list of keys that should be filtered out from the report and breadcrumb
     # metadata before sending them to Bugsnag
+    # @deprecated Use {#redacted_keys} instead
     # @return [Set<String, Regexp>]
     attr_accessor :meta_data_filters
 
+    # A list of keys that should be redacted from the report and breadcrumb
```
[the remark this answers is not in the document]
imjoehaines: Does the "(case insensitive)" imply that all regexes will be made case insensitive too? You can still do case sensitive matching with a regex like `/^ABC$/`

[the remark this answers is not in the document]
tomlongridge: How about just putting the sentence from your docs PR after the existing one:

> When adding strings, keys which are equal to the string (ignoring case) will be redacted. When adding regular expressions, any keys which match the regular expression will be redacted.

(Although I think it should be "keys **that** are equal".)

spec/cleaner_spec.rb
imjoehaines: Added in https://github.com/bugsnag/bugsnag-ruby/pull/703/commits/358f38194a9d36de4241040a07bb0d38fda3dda5

tomlongridge (approved):

PR #705, opened 2021-10-05
Add `Configuration#vendor_paths` to replace `Configuration#vendor_path`

imjoehaines:
## Goal

This PR adds `Configuration#vendor_paths` to replace `Configuration#vendor_path`

`vendor_paths` is an array of strings which will mark a stack frame as out of project if the frame's file path starts with any of the strings, after removing the project root (if a file is outside of the project root, it wouldn't be marked as in project in the first place)

For example, with this stacktrace:

```ruby
[
  "/foo/bar/app/models/user.rb:1:in `something'",
  "/foo/bar/abc/lib/ignore_me.rb:1:in `to_s'",
  "/foo/bar/xyz/lib/ignore_me.rb:1:in `to_s'",
]
```

and this configuration:

```ruby
Bugsnag.configure do |config|
  config.project_root = "/foo/bar"
  config.vendor_paths = ["abc", "xyz"]
end
```

only `/foo/bar/app/models/user.rb` will be marked as in project, as the other paths are matched by `vendor_paths`

Note that matches are done on entire directories, for example `abc` does not match `abc_xyz`

### DIFF
--- a/lib/bugsnag/configuration.rb
+++ b/lib/bugsnag/configuration.rb
@@ -155,9 +155,17 @@ module Bugsnag
     # @return [Integer]
     attr_reader :max_breadcrumbs
 
-    #
+    # @deprecated Use {vendor_paths} instead
     # @return [Regexp]
     attr_accessor :vendor_path
+
+    # An array of paths within the {project_root} that should not be considered
+    # as "in project"
+    #
+    # These paths should be relative to the {project_root}
+    #
+    # @return [Array<String>]
+    attr_accessor :vendor_paths
 
     # The default context for all future events
     # Setting this will disable automatic context setting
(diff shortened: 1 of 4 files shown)


lib/bugsnag/configuration.rb
```diff
@@ -155,10 +155,18 @@ class Configuration
...
 
-    #
+    # @deprecated Use {vendor_paths} instead
     # @return [Regexp]
     attr_accessor :vendor_path
 
+    # An array of paths within the {project_root} that should not be considered
+    # as "in project"
+    #
+    # These paths should be relative to the {project_root}
```
tomlongridge, line 165:
```suggestion
    # These paths should be relative to the {project_root} and will only match whole directory names
```

tomlongridge (approved):
LGTM - one suggested clarification on the comment-docs

PR #709, opened 2021-11-15
Tweak when we include the AR transaction callback rescuing

apalmblad:
## Goal

Bugsnag notifies when we see an exception in a transaction (or rollback) callback.  This is unnecessary as Rails now raises exceptions out of transaction callbacks.  This Bugsnag  behaviour means that certain errors are always sent to bugsnag instead of potentially handled elsewhere.  

Bugsnag behaviour can be seen here: https://github.com/apalmblad/bugsnag-ruby/blob/master/lib/bugsnag/integrations/rails/active_record_rescue.rb#L14

Rails has stopped swallowing exceptions in these callbacks, making in configurable in Rails 4, and then the default behaviour in Rails 5.
[Rails allowing exceptions to be raised](https://github.com/rails/rails/pull/16537)
[Rails 5 removing configuration for not raising callback exceptions](https://github.com/rails/rails/commit/07d3d402341e81ada0214f2cb2be1da69eadfe72)
## Design

To minimize changes needed, I consider the rails version and the various configuration values and conditionally include the file.

## Changeset

The rails integration behaviour was changed, and we stop including it.

## Testing

I've added a spec to check whether or not its included.

### DIFF
--- a/lib/bugsnag/integrations/railtie.rb
+++ b/lib/bugsnag/integrations/railtie.rb
@@ -70,8 +70,15 @@ module Bugsnag
       end
 
       ActiveSupport.on_load(:active_record) do
-        require "bugsnag/integrations/rails/active_record_rescue"
-        include Bugsnag::Rails::ActiveRecordRescue
+        if Rails::VESION::MAJOR == 4
+          if ActiveRecord::Base.errors_in_transactional_callbacks == :raise
+            require "bugsnag/integrations/rails/active_record_rescue"
+            include Bugsnag::Rails::ActiveRecordRescue
+          end
+        elsif Rails::VESION::MAJOR < 4
+          require "bugsnag/integrations/rails/active_record_rescue"
+          include Bugsnag::Rails::ActiveRecordRescue
+        end
       end
 
       ActiveSupport.on_load(:active_job) do
--- a/spec/integrations/active_record_spec.rb
+++ b/spec/integrations/active_record_spec.rb
@@ -1,8 +1,8 @@
 require 'spec_helper'
-
+require "bugsnag/integrations/rails/active_record_rescue"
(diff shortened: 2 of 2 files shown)


lib/bugsnag/integrations/railtie.rb
```diff
@@ -70,8 +70,15 @@ def event_subscription(event)
...
 
       ActiveSupport.on_load(:active_record) do
-        require "bugsnag/integrations/rails/active_record_rescue"
-        include Bugsnag::Rails::ActiveRecordRescue
+        if Rails::VESION::MAJOR == 4
+          if ActiveRecord::Base.errors_in_transactional_callbacks == :raise
+            require "bugsnag/integrations/rails/active_record_rescue"
+            include Bugsnag::Rails::ActiveRecordRescue
+          end
+        elsif Rails::VESION::MAJOR < 4
```
Jaacky, line 78:
```suggestion
        if Rails::VERSION::MAJOR == 4
          if ActiveRecord::Base.errors_in_transactional_callbacks == :raise
            require "bugsnag/integrations/rails/active_record_rescue"
            include Bugsnag::Rails::ActiveRecordRescue
          end
        elsif Rails::VERSION::MAJOR < 4
```

imjoehaines:
Thanks for the PR, @apalmblad!

I'll merge this in https://github.com/bugsnag/bugsnag-ruby/pull/717, where I've added tests to our Rails test apps to get better coverage of this behaviour

apalmblad:
Thanks for putting this in.

xander-jones:
Released in https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.24.2 🎉 , thanks for the contribution!

PR #710, opened 2021-11-18
resque: be more specific when handling ActiveJob wrapped jobs

isnotajoke:
## Goal

The Resque Bugsnag integration adds helpful context and payload information to reports from Resque tasks. A check within the integration attempts to unwrap `ActiveJob` jobs run via Resque. The check used to perform this unwrapping assumes that all jobs will run via `ActiveJob`, and causes errors when used with jobs other than `ActiveJob` whose first argument does not respond sensibly to `[]`. For example:

```
Error in internal notify block: no implicit conversion of String into Integer
```

after bubbling up to the top-level [Bugsnag notify method](https://github.com/bugsnag/bugsnag-ruby/blob/94da895ad148b9cd132099a6a0be5710a5f03e35/lib/bugsnag.rb#L89). This occurs before any Resque-specific context can be added to the report, and yields reports that are less useful than they could be. We want to prevent these errors so Resque reports have more uniform context.

## Design

The comment in the current implementation asserts an expectation about the payload of ActiveJob reports, but doesn't actually assert it or guard the following code by checking that assumption. It seemed reasonable to me to guard the error-producing conditional with an explicit assertion following the comment – restricting it to only those payloads for which it makes sense.

## Changeset

* Guard ActiveJob-specific Resque handling with an assertion that we're handling an ActiveJob report.

## Testing

- [ ] Confirmed in my environment that Resque jobs more reliably have Resque context associated with them
- [ ] Confirmed in my environment that "Error in internal notify block" errors (associated with this issue) are reduced/eliminated after this fix
- [ ] Confirmed in my environment that errors originating in ActiveJob tasks are properly unwrapped
- [ ] Added specs for this case

### DIFF
--- a/lib/bugsnag/integrations/resque.rb
+++ b/lib/bugsnag/integrations/resque.rb
@@ -49,7 +49,7 @@ module Bugsnag
 
         # when using Active Job the payload "class" will always be the Resque
         # "JobWrapper", not the actual job class so we need to fix this here
-        if metadata['args'] && metadata['args'][0] && metadata['args'][0]['job_class']
+        if class_name == 'ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper' && metadata['args'] && metadata['args'][0] && metadata['args'][0]['job_class']
           class_name = metadata['args'][0]['job_class']
           metadata['wrapped'] ||= class_name
         end


lib/bugsnag/integrations/resque.rb
```diff
@@ -49,7 +49,7 @@ def save
 
         # when using Active Job the payload "class" will always be the Resque
         # "JobWrapper", not the actual job class so we need to fix this here
-        if metadata['args'] && metadata['args'][0] && metadata['args'][0]['job_class']
+        if class_name == 'ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper' && metadata['args'] && metadata['args'][0] && metadata['args'][0]['job_class']
```
isnotajoke, line 52: There's a risk here that this may be too specific or accidentally a breaking change – either because it doesn't account for an ActiveJob-specific quirk, or because the existing code was coincidentally unwrapping exceptions from other frameworks. An alternative implementation would be wrapping this existing logic in a `begin/rescue`; it would still fail, but in a way that is less disruptive than the existing implementation. I'm open to going that way if it's preferred by maintainers

isnotajoke (commented):
Self-review. I still need to add specs and finish testing this; wanted to open the PR to get feedback on the implementation while I do that.

imjoehaines (commented):
Thanks for the PR, @isnotajoke

Sorry for this, we do have tests that run Resque jobs for real, but unfortunately the jobs aren't passed arguments so didn't trigger this bug 😕

https://github.com/bugsnag/bugsnag-ruby/blob/94da895ad148b9cd132099a6a0be5710a5f03e35/features/rails_features/integrations.feature#L72-L111

I'll make sure to change them to have arguments to check for this case and I can add some specs for this as well

imjoehaines:
I'll merge this in https://github.com/bugsnag/bugsnag-ruby/pull/711 where I've added & updated some tests

Thanks again!

xander-jones:
@isnotajoke this has been released in https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.24.1 thanks for the contribution! 🎉

PR #733, opened 2022-11-09
Prevent possible infinite loop in rake spec

imjoehaines:
## Goal

One of our Rake specs runs a command and expects it to report to Bugsnag. If this doesn't work for whatever reason, it gets stuck in an infinite loop. This is because `Queue#pop` blocks until there is something on the queue to pop, but if the command didn't work as expected there will never be anything added to the queue

We can't simply detect if the command worked or not as it's designed to crash (so will always have a failing exit code) and even if we ensured the output matched what we expect, it's still possible for this code to cause an infinite loop

I've changed the spec to use the non-blocking form of `Queue#pop` and make multiple attempts before failing, which will prevent it from looping infinitely

### DIFF
--- a/spec/integrations/rake_spec.rb
+++ b/spec/integrations/rake_spec.rb
@@ -44,15 +44,13 @@ describe "Bugsnag Rake integration" do
         res.status = 200
         res.body = "OK\n"
       end
-      Thread.new{ server.start }
+      Thread.new { server.start }
     end
 
     after do
       server.stop
       queue.clear
     end
-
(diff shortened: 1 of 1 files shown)


kattrali (commented):
Maybe it makes sense to replace or remove `let(:request)` (line 55)?

imjoehaines:
> Maybe it makes sense to replace or remove `let(:request)` (line 55)?

Yeah, definitely!

kattrali (approved):

PR #739, opened 2022-11-11
Update the "plain ruby" tests to use Maze Runner v7

imjoehaines:
## Goal

This PR migrates the plain ruby tests to use Maze Runner v7. This required some compatibility code in our custom steps as well as updating the feature files themselves. When all of our tests are running on v7, we can remove the compatibility code

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -38,19 +38,51 @@ def current_ip
   ip_list.captures.first
 end
 
-
-AfterConfiguration do |config|
-  install_fixture_gems
+def using_maze_runner_v7?
+  defined?(Maze)
 end
 
-Before do
-  Docker.compose_project_name = "#{rand.to_s}:#{Time.new.strftime("%s")}"
-  Runner.environment.clear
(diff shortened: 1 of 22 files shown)


features/support/env.rb
```diff
@@ -38,19 +38,51 @@ def current_ip
   ip_list.captures.first
 end
 
-
-AfterConfiguration do |config|
-  install_fixture_gems
+def using_maze_runner_v7?
+  defined?(Maze)
```
twometresteve, line 42: I guess this does the job, but you could also reference `Maze::VERSION` explicitly:

```
module Maze
  VERSION = '3.7.4'.freeze
end
```

and

```
module Maze
  VERSION = '7.5.1'
```

twometresteve (approved):
Just one suggestion.

features/support/env.rb
imjoehaines: Ah yeah, that would have been clearer! I think I'll leave it like this since after everything is upgraded it will get deleted anyway

twometresteve: Yep, no problem.

PR #740, opened 2022-11-11
Update remaining non-rails maze runner tests to use v7

imjoehaines:
## Goal

This updates the remaining non-rails maze runner tests to use v7:

- delayed_job
- mailman
- que
- rack
- rake
- sidekiq

Most of the changes are but a couple of the commands that were being run also had to change as Maze Runner doesn't run commands in the background as of v3.2.0: https://github.com/bugsnag/maze-runner/commit/78425e016960ce50cab593991a03187d9e79f067

### DIFF
--- a/features/fixtures/delayed_job/app/Rakefile
+++ b/features/fixtures/delayed_job/app/Rakefile
@@ -20,9 +20,9 @@ namespace :delayed_job_tests do
 end
 
 def run_delayed_job_test(command)
-  fork do
-    system("rake jobs:work")
-  end
+  # queue the job with rails' runner
   system("rails runner #{command}")
-  Process.wait
+
+  # run the queued jobs and exit
+  system("rake jobs:workoff")
 end
--- a/features/que.feature
+++ b/features/que.feature
@@ -2,9 +2,9 @@ Feature: Errors are delivered to Bugsnag from Que
 
 Scenario: Que will deliver unhandled errors
   Given I run the service "que" with the command "bundle exec ruby app.rb unhandled"
-  And I run the service "que" with the command "bundle exec que ./app.rb"
-  And I wait to receive a request
-  Then the request is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier"
+  And I run the service "que" with the command "timeout 5 bundle exec que ./app.rb"
+  And I wait to receive an error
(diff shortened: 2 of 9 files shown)


features/fixtures/delayed_job/app/Rakefile
```diff
@@ -20,9 +20,9 @@ namespace :delayed_job_tests do
...
 def run_delayed_job_test(command)
-  fork do
-    system("rake jobs:work")
-  end
+  # queue the job with rails' runner
   system("rails runner #{command}")
-  Process.wait
+
+  # run the queued jobs and exit
+  system("rake jobs:workoff")
```
imjoehaines, line 27: The `workoff` command runs until there are no more queued jobs, then exits

features/que.feature
```diff
@@ -2,9 +2,9 @@ Feature: Errors are delivered to Bugsnag from Que
 
 Scenario: Que will deliver unhandled errors
   Given I run the service "que" with the command "bundle exec ruby app.rb unhandled"
-  And I run the service "que" with the command "bundle exec que ./app.rb"
-  And I wait to receive a request
-  Then the request is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier"
+  And I run the service "que" with the command "timeout 5 bundle exec que ./app.rb"
```
imjoehaines, line 5: `timeout` is necessary here because Que doesn't have an equivalent of Delayed Job's `workoff` — as this is run in the foreground now, it would be stuck infinitely

features/sidekiq.feature
```diff
@@ -1,9 +1,9 @@
 Feature: Bugsnag raises errors in Sidekiq workers
 
 Scenario: An unhandled RuntimeError sends a report
-  Given I run the service "sidekiq" with the command "bundle exec rake sidekiq_tests:unhandled_error"
-  And I wait to receive a request
-  Then the request is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier"
+  Given I run the service "sidekiq" with the command "timeout 5 bundle exec rake sidekiq_tests:unhandled_error"
```
imjoehaines, line 4: `timeout` is needed here for the same reason as Que

twometresteve (approved):

PR #742, opened 2022-11-11
Run Rails integrations tests with Maze Runner v7

imjoehaines:
## Goal

This PR migrates the last set of tests over to Maze Runner v7 🎉

I'll cleanup the Maze Runner v3/v7 compatibility stuff separately, to keep this PR small

### DIFF
--- a/features/rails_features/integrations.feature
+++ b/features/rails_features/integrations.feature
@@ -7,10 +7,10 @@ Background:
 
 @rails_integrations
 Scenario: Delayed job
-  When I run the "jobs:work" rake task in the rails app
-  And I run "User.new.delay.raise_the_roof" with the rails runner
-  And I wait to receive a request
-  Then the request is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier"
+  When I run "User.new.delay.raise_the_roof" with the rails runner
+  And I run the "jobs:workoff" rake task in the rails app
+  And I wait to receive an error
+  Then the error is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier" notifier
   And the event "unhandled" is true
   And the event "severity" equals "error"
   And the event "context" equals "User#raise_the_roof"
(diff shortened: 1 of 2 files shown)


features/rails_features/integrations.feature
```diff
@@ -71,10 +71,10 @@ Scenario: Rake
 
 @rails_integrations
 Scenario: Resque (no on_exit hooks)
-  When I run "bundle exec rake resque:work" in the rails app
-  And I run "Resque.enqueue(ResqueWorker, 123, %(abc), x: true, y: false)" with the rails runner
-  And I wait to receive a request
-  Then the request is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier"
+  When I run "Resque.enqueue(ResqueWorker, 123, %(abc), x: true, y: false)" with the rails runner
+  And I run "timeout --signal QUIT 5 bundle exec rake resque:work" in the rails app
```
imjoehaines, line 75: [`timeout` uses the `TERM` signal by default](https://www.man7.org/linux/man-pages/man1/timeout.1.html), which causes Resque to report an error (and the test to fail as it gets 2 events)

[Resque gracefully shutsdown when given the `QUIT` signal](https://github.com/resque/resque#signals), so we have to tell `timeout` to do that instead

features/rails_features/integrations.feature
```diff
@@ -7,10 +7,10 @@ Background:
 
 @rails_integrations
 Scenario: Delayed job
-  When I run the "jobs:work" rake task in the rails app
-  And I run "User.new.delay.raise_the_roof" with the rails runner
-  And I wait to receive a request
-  Then the request is valid for the error reporting API version "4.0" for the "Ruby Bugsnag Notifier"
+  When I run "User.new.delay.raise_the_roof" with the rails runner
+  And I run the "jobs:workoff" rake task in the rails app
```
imjoehaines, line 11: The order of these had to change so that the job gets queued first and the job worker runs second, otherwise the job worker command will block the job from being queued in the first place

PR #745, opened 2022-11-15
Add FeatureFlagDelegate class

imjoehaines:
## Goal

Add the `FeatureFlagDelegate` class, which is used to manipulate stored feature flags in much the same way as [the `MetadataDelegate`](https://github.com/bugsnag/bugsnag-ruby/blob/caf3ecc9aa812af4df0275861a739da297e83398/lib/bugsnag/utility/metadata_delegate.rb) works for metadata values

### DIFF
--- a/lib/bugsnag/utility/feature_flag_delegate.rb
+++ b/lib/bugsnag/utility/feature_flag_delegate.rb
@@ -1,0 +1,77 @@
+module Bugsnag::Utility
+  # @api private
+  class FeatureFlagDelegate
+    def initialize
+      # feature flags are stored internally in a hash of "name" => <FeatureFlag>
+      # we don't use a Set because new feature flags should overwrite old ones
+      # that share a name, but FeatureFlag equality also uses the variant
+      @storage = {}
+    end
+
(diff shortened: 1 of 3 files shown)


lib/bugsnag/utility/feature_flag_delegate.rb
```diff
@@ -0,0 +1,77 @@
...
+      @storage = {}
+    end
+
+    # Add a feature flag with the given name & variant
+    #
+    # @param name [String]
+    # @param variant [String, nil]
+    # @return [void]
+    def add(name, variant)
+      @storage[name] = Bugsnag::FeatureFlag.new(name, variant)
```
kattrali, line 17: potentially silly thing but if `nil` is passed as the name, it will be used as the key and may be a problem later.

kattrali (approved):

lib/bugsnag/utility/feature_flag_delegate.rb
kattrali: Sounds good :+1:

PR #755, opened 2022-11-23
Scope feature flags to the current thread

imjoehaines:
## Goal

This allows users to add feature flags to Bugsnag whenever they interact with their feature flag service, e.g.

```ruby
flag = get_flag_from_service('new-login-page')

if flag.enabled?
  Bugsnag.add_feature_flag('new-login-page')
  # render new login page
else
  Bugsnag.clear_feature_flag('new-login-page')
  # render old login page
end
```

Before this PR, the above snippet would affect all web requests/queued jobs because `Bugsnag.add_feature_flag` would write to a global set of flags (similar to `Bugsnag.add_metadata` vs `event.add_metadata`)

With this PR, feature flags are stored in a thread local (just like breadcrumbs), meaning each web request or queued job will have a separate set of flags

This is important because feature flags will usually be user-specific, e.g. to allow rolling out features to small groups of users at a time. With the previous implementation this was still possible but meant relying on adding all flags for the user at once in an `on_error` callback, whereas the new implementation allows for adding flags individually when they are used

## Changeset

- Store the `Configuration`'s `FeatureFlagDelegate` in a thread local
- Moved feature flag API into a `FeatureDataStore` module to allow all 3 places that implement the API to share the same code & docs (`Bugsnag`, `Configuration` & `Event`/`Report`)
- Updated Maze Runner tests to reflect the new changes — these previously relied on some flags being global

### DIFF
--- a/features/fixtures/rack/app/app.rb
+++ b/features/fixtures/rack/app/app.rb
@@ -2,6 +2,8 @@ require 'bugsnag'
 require 'rack'
 require 'json'
 
+$clear_all_flags = false
+
 Bugsnag.configure do |config|
   config.api_key = ENV['BUGSNAG_API_KEY']
   config.endpoint = ENV['BUGSNAG_ENDPOINT']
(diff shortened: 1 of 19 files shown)


lib/bugsnag.rb
```diff
@@ -595,6 +562,13 @@ def unwrap_bundler_exception(exception)
 
       unwrapped.cause
     end
+
+    # Expose the feature flag delegate for {Bugsnag::Utility::FeatureDataStore}
+    #
```
tomlongridge, line 567: Should we annotate with `@api private` as per the declaration in `configuration`?

features/fixtures/rack/app/app.rb
```diff
@@ -2,19 +2,41 @@
...
   config.endpoint = ENV['BUGSNAG_ENDPOINT']
 
   if ENV.key?('BUGSNAG_METADATA_FILTERS')
     config.meta_data_filters = JSON.parse(ENV['BUGSNAG_METADATA_FILTERS'])
   end
+
+  config.add_feature_flag(
+    'this flag is added outside of a request and so should never normally be visible',
+    'if an event was reported in the global scope, this flag would be attached to it'
+  )
```
tomlongridge, line 18: I'm assuming this now won't be the case once this PR has been rebased from the #754 changes? I would expect global flags to appear in all events. If that's not the case, then we probably need to discuss this.

lib/bugsnag.rb
imjoehaines: It doesn't need to be because it's actually private: https://github.com/bugsnag/bugsnag-ruby/blob/4b4203ca81cd048a60d7ac94c72812075f9aacf6/lib/bugsnag.rb#L435

features/fixtures/rack/app/app.rb
imjoehaines: This branch was already based on #754. There's no such thing as a global feature flag anymore

You can make a flag always show up by adding it in an `on_error` callback, but any you add using `Bugsnag` or `Bugsnag::Configuration` are stored in a thread local so they are request/job specific

tomlongridge (approved):
Approving on the basis of further work on global feature flags.

PR #762, opened 2022-12-19
Fix mailman fixture

imjoehaines:
## Goal

A newly released version of the `mail` gem broke the fixture by requiring too new of a Ruby version (via one of its dependencies `date`)


twometresteve (approved):
Pending CI

Cawllec (approved):
Based on CI passing

PR #765, opened 2022-12-20
Add Ruby 3.2 to CI

imjoehaines:
## Goal

Update CI to use Ruby 3.2

## Changeset

- `maze-runner.yml`: use the docker tag `3.2-rc` where we're currently running 3.1 — there's no `3.2` tag yet so this will need updating when 3.2 is officially released
- `test-package.yml`: add 3.2 as a version to run specs with (setup-ruby understands that without a RC suffix)


twometresteve (approved):
LGTM pending CI

PR #767, opened 2022-12-28
fix typo in CHANGELOG

timdiggins:
## Goal

There was a typo in the CHANGELOG which lists the most recent release as v6.26.0 - when it was actually v6.25.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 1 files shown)


imjoehaines:
Nice spot, @timdiggins!

I'm going to close this as I've already fixed it in the `next` branch (https://github.com/bugsnag/bugsnag-ruby/pull/766), which will get merged soon when I make the v6.26.0 release for real

PR #769, opened 2023-01-04
Fix typo in version number

das-salzamt:


johnkiely1:
Hi @das-salzamt. 

Thanks for raising, it does look like this is a duplicate of https://github.com/bugsnag/bugsnag-ruby/pull/767 which will be released shortly. Going to close this.

PR #771, opened 2023-01-04
Use docker compose exec to run queue commands in maze runner tests

imjoehaines:
## Goal

Switch to using `docker compose exec` to run queue commands in the Maze Runner tests

This is faster and more reliable than the `timeout` based approach we're currently using

### DIFF
--- a/features/rails_features/integrations.feature
+++ b/features/rails_features/integrations.feature
@@ -1,14 +1,10 @@
 Feature: App type is set correctly for integrations in a Rails app
 
-Background:
-  Given I start the rails service
-  And I run the "db:prepare" rake task in the rails app
-  And I run the "db:migrate" rake task in the rails app
-
 @rails_integrations
 Scenario: Delayed job
+  Given I start the rails service with the database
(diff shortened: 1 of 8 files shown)


features/rails_features/integrations.feature
```diff
@@ -1,14 +1,10 @@
 Feature: App type is set correctly for integrations in a Rails app
 
-Background:
-  Given I start the rails service
-  And I run the "db:prepare" rake task in the rails app
-  And I run the "db:migrate" rake task in the rails app
```
imjoehaines, line 6: This background no longer works as the environment variable `ACTIVE_JOB_QUEUE_ADAPTER` is set too late now — `docker compose exec` re-uses the already running container, so environment variables must be set before starting the container

imjoehaines: Most of these tests didn't need the database setup anyway — only Que & Delayed Job rely on it

axljoker03 (approved):

Cawllec (approved):

PR #772, opened 2023-01-12
`bundle install` with `--with` and `--binstubs` flags has been deprecated

sato11:
Following the getting started in CONTRIBUTING.md outputs deprecation messages:

```
[DEPRECATED] The `--with` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local with 'test sidekiq'`, and stop using this flag
[DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs --all`
```

This commit updates the section accordingly.

Also, since it encourages to generate binstubs, incentivize using bin/* over bundle exec in development.

## Goal
- keep development onboarding free of warnings
- comply with bundler instructions

## Changeset
Documentation only.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,12 +17,14 @@ Thank you!
 
 * Install the test dependencies
     ```
-    bundle install --with test sidekiq --binstubs
+    bundle config set --local with 'test sidekiq'
+    bundle install
+    bundle binstubs --all
     ```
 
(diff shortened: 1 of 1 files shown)


johnkiely1:
Thanks for the PR, we will look to review when priorities allow

imjoehaines (approved):
Thanks @sato11!

PR #773, opened 2023-01-12
Replace `Bundler.with_clean_env` with `Bundler.with_unbundled_env`

sato11:
Running spec outputs the following deprecation message:

```
[DEPRECATED] `Bundler.with_clean_env` has been deprecated in favor of `Bundler.with_unbundled_env`. If you instead want the environment before bundler was originally loaded, use `Bundler.with_original_env` (called at /[abridged]/spec/integrations/logger_spec.rb:88)
```

This commit switches them accordingly and make the test suite free of bundler warnings.

## Design
The message says we have another option, `Bundler.with_original_env`, which I did not opt for as simply replacing with `with_unbundled_env` seems alright for tests to pass.

## Testing
The CI workflow will let us know that bin/rake passes.

### DIFF
--- a/spec/integrations/logger_spec.rb
+++ b/spec/integrations/logger_spec.rb
@@ -19,7 +19,7 @@ describe 'Configuration.logger' do
     def run_app(name)
       out_reader, out_writer = IO.pipe
       Dir.chdir(File.join(File.dirname(__FILE__), "../fixtures/apps/#{name}")) do
-        Bundler.with_clean_env do
+        Bundler.with_unbundled_env do
           pid = Process.spawn('bundle install',
                               out: out_writer.fileno,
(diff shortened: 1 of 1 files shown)


johnkiely1:
Thanks for the PR, we will review as soon as priorities allow.

imjoehaines:
Hey @sato11, thanks for the PR

Unfortunately this isn't mergeable as `with_unbundled_env` doesn't exist in the old Bundler versions that we have to use when testing against Ruby 1.9

These tests are on my list to rewrite as they're quite slow and can break easily, so hopefully this deprecation notice won't be around that much longer anyway

sato11:
That's a good point. Thanks for maintaining 👍

PR #776, opened 2023-02-13
Allow partial strings on Rails `filter_parameters` for `meta_data_filters`

yykamei:
## Goal

The Recent Rails template will generate `filter_parameters` with *partial* strings, such as `:passw` and `:_key`. I think this library should also allow such keywords to filter out, following the Rails template.

https://github.com/rails/rails/blob/f838a7421228204bbc1e012ff9a3801ed598da80/railties/lib/rails/generators/rails/app/templates/config/initializers/filter_parameter_logging.rb.tt#L6-L8

This patch might be a breaking change because it will filter more parameters than before, but to exclude them would be better than sending probably sensitive parameters as they are.

## Design

I just removed `\A` and `\z` from the `Regexp` construction on a railtie as well as adding `i` to allow any cases to be matched.

## Changeset

* Remove `\A` and `\z`, and add `i` to `Regexp` initialization on railtie

## Testing

I updated `Rails.application.config.filter_parameters` inside `features` to be `:passw` from `:password`. It would be good if the End to end tests pass.

### DIFF
--- a/example/rails-42/config/initializers/filter_parameter_logging.rb
+++ b/example/rails-42/config/initializers/filter_parameter_logging.rb
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [:passw]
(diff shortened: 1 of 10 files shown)


johnkiely1:
Hi @yykamei, Thanks for the PR, we will review when priorities allow.

yykamei:
Hi @johnkiely1, I'm curious about the progress of this pull request. I'm ready to update any changes if they're necessary. Thank you!

clr182:
Hi @yykamei, due to the prioritization of other tasks we do not have an ETA on the review of this PR. We'd be happy to let you know once we have more information.

yykamei:
Thank you for your reply.

PR #783, opened 2023-06-24
docs: remove bugsnag-em from the docs

tnir:
## Goal

`bugsnag` gem 5.x is no more used and so `bugsnag-em` should be unreferenced.

## Design

As per https://github.com/bugsnag/bugsnag-em/pull/5, we should remove references for `bugsnag-em` gem.

## Changeset

- Remove an entry about bugsnag-em from the docs.

## Testing

n/a


tnir:
@imjoehaines Status?

imjoehaines (approved):

PR #785, opened 2023-07-07
Add support for Sidekiq 7

stevenharman:
Sidekiq 7 has a new internal structure and middleware is now expected to include a module, depending on client vs server middleware. See: https://github.com/sidekiq/sidekiq/blob/main/docs/middleware.md

## Goal

Support for Sidekiq 7, which has some new internals, including a slightly different middleware API. While we don't currently any of the things offered by the new API, this will make sure we don't break on accident.

## Design

Only include the new middleware module if it's been defined, which will only happen for Sidekiq 7+.

### DIFF
--- a/.github/workflows/maze-runner.yml
+++ b/.github/workflows/maze-runner.yml
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


imjoehaines (approved):
Thanks @stevenharman!

PR #797, opened 2023-10-19
Fix /examples filepaths on Windows

johnnyshields:
On Windows, file names cannot contain `:` (colon) character. This PR simply renames the `example/rails-51/lib/tasks/resque:setup.rake` file to `resque_setup.rake`. Since this is just an example file, it doesn't actually affect the gem.


clr182:
Hi @johnnyshields

PR #798, opened 2023-10-19
Add "sort" field to Mongo breadcrumb

johnnyshields:
## Goal

Add "sort" field to breadcrumb for MongoDB queries. Sort can have significant bearing on index selection/performance, and its very useful to see.

## Design

The implementation mirrors how the "filter" field is captured today. Unlike filters, we don't need to mask the values  because they can only be 1 or -1.

## Changeset

The Mongo breadcrumb functionality only.

This also includes the changes from https://github.com/bugsnag/bugsnag-ruby/pull/797 b/c I cannot commit from my Windows machine otherwise.


clr182:
Hi @johnnyshields 

johnnyshields:
Here is evidence of this working:

image

PR #799, opened 2023-10-19
Mongo filter array length

johnnyshields:
## Goal

In Mongo, it is very common to do an `"$in": [ some long array ]` to avoid N+1 queries. Here's how this looks in the breadcrumb:
image

This is a problem because:
1. Its a ton of text
2. It's difficult to know how many members the array actually has.

## Design

To fix this, a Array which (a) has more than 3 members and (b) does not contain nested Arrays or Hashes will be summarized as `["LENGTH=42"]` (e.g. if it has 42 members)

## Changeset

The breadcrumb generator for MongoDB

## Testing

Tested in RSpec and in production by my company.


clr182:
Hi @johnnyshields 

PR #801, opened 2023-10-19
Mongo breadcrumb: Do not mask filter value nil

johnnyshields:
## Goal

In Mongo queries, the value `nil` is a special in that it can impact the behavior of sparse indexes. Rather than masking it with "?" we should indicate that it is nil. There is no risk of this causing data leakage to the logs.

## Design

This is implemented in the filter masking function.

## Changeset

Only changes the Mongo breadcrumb functionality.

This also includes the changes from:
- https://github.com/bugsnag/bugsnag-ruby/pull/797 b/c I cannot commit from my Windows machine otherwise.
- https://github.com/bugsnag/bugsnag-ruby/pull/798 to avoid conflicts
- https://github.com/bugsnag/bugsnag-ruby/pull/799 to avoid conflicts

## Testing

Tested in RSpec and by [TableCheck](www.tablecheck.com/en/join) in production.


johnnyshields:
@mclack and Bugsnag team can you please take a look at my PRs?

clr182:
Hi @johnnyshields 

This is still on our backlog and we'll be sure to keep you updated once we have an update. Is data leakage an issue that you are facing consistently while using bugsnag-ruby ?

PR #806, opened 2023-12-18
Redact URLs in automatic Rails breadcrumbs

imjoehaines:


tomlongridge (approved):

lib/bugsnag/integrations/railtie.rb
```diff
@@ -36,6 +37,12 @@ def event_subscription(event)
             # the connection ID is the object_id of the connection object
             filtered_data[:connection_id] = data[:connection].object_id
           end
+
+        when "start_processing.action_controller"
+          filtered_data[:path] = Bugsnag.cleaner.clean_url(data[:path]) if data.key?(:path)
+
+        when "redirect_to.action_controller"
+          filtered_data[:location] = Bugsnag.cleaner.clean_url(data[:location]) if data.key?(:location)
```
timdiggins, line 45: This causes an error to surface in the app if the `data[:location]` (ie. the redirect_to *url*) is a badly formed, but actually in practice working URL, such as: `https://example.com/some path with spaces` -- browsers (and even rack:test) will automatically convert this to `https://example.com/some path%20with%20spaces`  but `URI("https://example.com/some path with spaces")` (as used at https://github.com/bugsnag/bugsnag-ruby/blob/master/lib/bugsnag/cleaner.rb#L30) will raise URI::InvalidURIError

imjoehaines: Thanks, @timdiggins! I'll try to get a release out with a fix early next week

Issue #810, opened 2024-01-11
Bugsnag instrumentation causes URI: on redirect to badly formed (but working) url


timdiggins:
### Describe the bug

In v6.2.1, there is an addition to redact potential secrets from the location of redirects. This uses `Bugsnag.cleaner.clean_url` on the location which in turn uses `URI()` - this is more sensitive than browsers and will raise an exception on some RFC 3986 non-compliant but in practice actually working URLs (e.g. with spaces). 

https://github.com/bugsnag/bugsnag-ruby/pull/806/files#r1449084164

As a result this addition can cause errors in specs and in practice.

### Steps to reproduce

create a route & controller action that uses `redirect_to "http://example.com/some spaces"`
exercise this route & controller action in a test and/or in reality with Bugsnag v6.26.0 -- will work successfully

exercise this route & controller action in a test and/or in reality with Bugsnag v6.26.1 -- will raise an error.

### Suggestion

While it's not good practice in client code to have these non-compliant urls in any repo, I don't think Bugsnag should be the point of enforcement.  As far as possible Bugsnag should not introduce any new exceptions - maybe worth rescueing any exception during a clean and skip the clean (return input) if the URI raises an error

### Environment

* Bugsnag version: v6.26.1
* Ruby version: 3.0.6
* Bundle version: 2.2.33
* Integration framework version:
    * Rack: 2.2.8
    * Rails: 6.1.7.6

keegnotrub:
Just a note that I've also experienced issues with this, even after the recent v6.26.2 release, causing us to need to revert back to v6.26.0. Checking our code, links like the following seem to fail still in the most recent release:

```ruby
cleaner = Bugsnag::Cleaner.new(Bugsnag.configuration)
cleaner.clean_url("mailto:?subject=Hello")
```

clr182:
Hi @keegnotrub, 

This issue has been addressed in [v6.26.3](https://github.com/bugsnag/bugsnag-ruby/releases/tag/v6.26.3) of the bugsnag-ruby notifier. If after upgrade you still see issues, please let us know and we can investigate further.

PR #811, opened 2024-01-16
Fix unhandled `URI::InvalidURIError` in `Cleaner#clean_url`

imjoehaines:
## Goal

As reported in #810, we don't handle invalid URLs in clean_url so raise if the URL isn't valid

If the URL doesn't have a query string then it's not important that it fails validation and so we can return it as-is

If the URL does have a query string and cannot be parsed, we can't redact parameters individually and so redact the whole query string instead — this way the URL could still be useful and we don't risk leaking sensitive data


tomlongridge (approved):

stevenharman:
Ooof, this was also causing a memory leak on our app due to the way older Rails (6.1 and 7.0, at least) manage their `ActiveSupport::Subscriber` array.

Basically, each `ActiveSupport::Notifications.subscribe` block pushes an `ActiveSupport::Notifications::Event` onto the `ActiveSupport::Subscriber#event_stack`. Those events are then popped off in the reverse order (last-in, first out) and processed. The processing of such `Event`s is the mechanism used by Bugsnag to scrub those URLs/paths. When that mechanism raises an error, as was happening here, in an inner `Event`, the outer `Event`s are left on the `event_stack`. Forever. Thus leaking memory.

This PR fixes that, and the entire `event_stack` has been [removed in Rails 7.1](https://github.com/rails/rails/pull/43390). Phew!

PR #813, opened 2024-01-23
Fix `Cleaner#clean_url` with mailto links

imjoehaines:
## Goal

mailto links don't have a `query` property, instead they use `headers`, e.g.:

```ruby
uri = URI('mailto:hello@example.com?subject=hi')
uri.headers #=> [['subject', 'hi']]
```

The current `Cleaner#clean_url` fails on mailto links and wouldn't filter the headers if it did succeed. This PR updates it to handle filtering the headers in mailto links

### DIFF
--- a/lib/bugsnag/cleaner.rb
+++ b/lib/bugsnag/cleaner.rb
@@ -30,16 +30,18 @@ module Bugsnag
 
       begin
         uri = URI(url)
+
+        if uri.is_a?(URI::MailTo)
+          clean_mailto_url(url, uri)
+        else
+          clean_generic_url(url, uri)
+        end
       rescue URI::InvalidURIError
         pre_query_string, _query_string = url.split('?', 2)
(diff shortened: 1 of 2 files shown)


tomlongridge (approved):
Changelog entry needed, but fix LGTM

lib/bugsnag/cleaner.rb
```diff
@@ -30,16 +30,18 @@ def clean_url(url)
...
 
-        return "#{pre_query_string}?#{FILTERED}"
-      end
-
-      if uri.is_a?(URI::MailTo)
-        clean_mailto_url(url, uri)
-      else
-        clean_generic_url(url, uri)
+        "#{pre_query_string}?#{FILTERED}"
+      rescue StandardError
```
timdiggins, line 43: Wonderful! Thanks for this.

PR #819, opened 2024-03-25
Only notify exception which are not rescued

ChrisBr:
## Goal

When specifying an `rescue_responses` in Rails, I would expect Bugsnag to not report this error.

```
config.action_dispatch.rescue_responses["FrozenRecord::RecordNotFound"] = :not_found
```

This is happening at the moment because the bugsnag Rack middleware runs before Rails [ShowException middleware](https://github.com/rails/rails/blob/61a3e61e7761bd0a66073f42d986286300cdce6b/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L31-L45).

I also don't want to ignore the error altogether because e.g. I want to get notified when it's triggered from a background job.

## Design

To be discussed

## Testing

n/a

### DIFF
--- a/lib/bugsnag/integrations/rack.rb
+++ b/lib/bugsnag/integrations/rack.rb
@@ -50,13 +50,16 @@ module Bugsnag
       begin
         response = @app.call(env)
       rescue Exception => raised
-        # Notify bugsnag of rack exceptions
-        Bugsnag.notify(raised, true) do |report|
-          report.severity = "error"
-          report.severity_reason = {
-            :type => Bugsnag::Report::UNHANDLED_EXCEPTION_MIDDLEWARE,
-            :attributes => Bugsnag::Rack::FRAMEWORK_ATTRIBUTES
-          }
+
+        if notify?
+          # Notify bugsnag of rack exceptions
+          Bugsnag.notify(raised, true) do |report|
+            report.severity = "error"
+            report.severity_reason = {
+              :type => Bugsnag::Report::UNHANDLED_EXCEPTION_MIDDLEWARE,
+              :attributes => Bugsnag::Rack::FRAMEWORK_ATTRIBUTES
+            }
+          end
         end
 
         # Re-raise the exception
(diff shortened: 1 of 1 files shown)


ChrisBr:
Let me know if this is something you consider or if you recommend a different configuration to achieve the same / similar goal. If this is something you would consider accepting, I'm happy to add tests and / or adapt the implementation.

mclack:
Hi @ChrisBr 

Apologies for the delay in responding to this.

You could implement an [on_error callback](https://docs.bugsnag.com/platforms/ruby/rails/customizing-error-reports/#on-error-callbacks) within your application to ignore `FrozenRecord::RecordNotFound` based on certain conditions. For example, something similar to the following could be useful:

```
Bugsnag.configure do |config|
  config.add_on_error(proc do |event|

    # ignore events from 'FrozenRecord::RecordNotFound' errors that happen in
    # HTTP requests
    if event.request && event.original_error.class.name == "FrozenRecord::RecordNotFound"
      false

    end
  end)
end
```

As you mention, it's likely that you wouldn't want to ignore all instances of this error, and there could still be cases where being notified of a rescued error could be valuable. Perhaps you could test the implementation above and try adding additional conditions to get the behaviour you desire.

If this doesn't cover what you want or you're having any difficulties in further configuring the callback, please feel free to provide some more detail on what you'd like to achieve and we'd be more than happy to help. As this is a public thread, you may prefer to open a support ticket with us directly by contacting support@bugsnag.com.

ChrisBr:
Thanks for the reply @mclack, I will try this.

My concern was more that I would expect this to be the default, no?

mclack:
Hi @ChrisBr 

We believe the safer option is to report these exceptions by default, and give users the option to configure whether they'd like to ignore them or not. It's possible that these exceptions could be thrown for reasons other than someone hitting a route with a non-existent ID, and reporting these could give valuable insight into why and how often they are occurring.

As this is in line with BugSnag's expected behaviour, we are going to close this issue now. As previously mentioned, please feel free to reach out to support@bugsnag.com if you'd like any assistance with ignoring these exceptions using an [on-error callback](https://docs.bugsnag.com/platforms/ruby/rails/customizing-error-reports/#on-error-callbacks).

PR #821, opened 2024-05-01
Include warden scope in user info

### LINKED ISSUE #777: Set warden scope on user data
### Description
We had removed any user PII from our Bugsnag integration in a large Rails app but we noticed that the default `user` tab data may still get set (with clear user data) when our `before_bugsnag_notify` controller method wouldn't apply, e.g. `ActionController::InvalidAuthenticityToken` error. 

While we can remove the user PII entirely, we'd rather just include the `id` and a link to our admin for that user, just as we do with our current `before_bugsnag_notify` controller method concern. The issue is that we have `User` and `AdminUser` models using Devise and with Bugsnag's `WardenUser` middleware it does not provide the warden scope.

https://github.com/bugsnag/bugsnag-ruby/blob/e5afde582907bd902b1161b99792522098742694/lib/bugsnag/middleware/warden_user.rb#L25-L30

So if we had an `id` of 2, we wouldn't know if it's a `User` or `AdminUser` without the warden scope.

**Describe the solution you'd like**

Include a `scope` key in the `user` object with the warden scope value so its available through the event user data documented in: https://docs.bugsnag.com/platforms/ruby/rails/customizing-error-reports/#user as part of `event.user[:scope]`, for example:

```
puts event.user[:id], event.user[:scope]
> 2
> :admin_user
```

With this new key, we could use that in an on-error callback https://docs.bugsnag.com/platforms/ruby/rails/customizing-error-reports/#on-error-callbacks in our Bugsnag configuration vs having in each controller (through a concern).

On a related note, are methods like `current_user` available from an on-error callback?

javierjulio:
## Goal

Closes #777

The default user metadata includes `id` but without the Warden scope, if you have multiple user types, for example: `User` and `AdminUser`, it wouldn't be possible to know who's who when removing the default fields which are PII.

## Design

Simple approach as Warden scope is already used to determine the user object in this case, so the objective is to just include the scope value as another user field.

## Changeset

Just adds a `warden_scope` key to the user object.

### DIFF
--- a/lib/bugsnag/middleware/warden_user.rb
+++ b/lib/bugsnag/middleware/warden_user.rb
@@ -21,7 +21,7 @@ module Bugsnag::Middleware
           best_scope = warden_scopes.include?("user") ? "user" : warden_scopes.first
 
           # Extract useful user information
-          user = {}
+          user = { :warden_scope => best_scope }
           user_object = env["warden"].user({:scope => best_scope, :run_callbacks => false}) rescue nil
           if user_object
             # Build the user info for this scope


imjoehaines (approved):
Thank you, @javierjulio!

Issue #822, opened 2024-05-16
warning: loading in progress, circular require considered harmful


ArtemDordovskyi:
### Describe the bug
I see warnings about circular require every time I run rake tasks.

### Steps to reproduce
1. Create rails application
```
rails new test-bugsnag
cd test-bugsnag
```
2. Update rubygems
```
gem update --system
```
2. Add bugsnag to gems 
```
bundle add bugsnag
```
3. Add `$VERBOSE = true` to `./config/application.rb` config file at the top
```
# ./config/application.rb
require_relative "boot"

$VERBOSE = true

require "rails/all"

# Require the gems listed in Gemfile, including any gems
# you've limited to :test, :development, or :production.
Bundler.require(*Rails.groups)
```
4. Then run command and see warning
```
bundle exec rake -T
```

### Environment
* Bugsnag version: 6.26.4
* Ruby version: 3.2.1
* Bundle version: 2.4.5
* Integration framework version:
    * Rack: 3.0.11
    * Rails: 7.1.3.2
    * Rake: 13.2.1

### Example Repo

- [x] Create a minimal repository that can reproduce the issue
- [x] Link to it here: https://github.com/ArtemDordovskyi/test-bugsnag

<details><summary>Error messages:</summary>

```
/home/user/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30: warning: /home/user/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30: warning: loading in progress, circular require considered harmful - /home/user/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/bugsnag-6.26.4/lib/bugsnag.rb
        from /home/user/.rbenv/versions/3.2.1/bin/bundle:25:in  `<main>'
        from /home/user/.rbenv/versions/3.2.1/bin/bundle:25:in  `load'
        from /home/user/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/bundler-2.4.5/exe/bundle:33:in  `<top (required)>'
[...]

PR #828, opened 2024-06-12
Fix circular require warning

imjoehaines:
## Goal

This PR fixes a circular require warning caused by several files requiring the main `bugsnag.rb` file

I've also gone through and removed other duplicate requires, consolidating them into `bugsnag.rb`

See https://github.com/bugsnag/bugsnag-ruby/issues/822

### DIFF
--- a/lib/bugsnag.rb
+++ b/lib/bugsnag.rb
@@ -1,5 +1,10 @@
 require "rubygems"
 require "thread"
+require "set"
+require "json"
+require "uri"
+require "socket"
+require "logger"
 
 require "bugsnag/version"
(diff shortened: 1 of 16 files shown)


tomlongridge:
Is it worth considering a rake task to check for this in future?

imjoehaines:
> Is it worth considering a rake task to check for this in future?

I think the best approach would be to fail the Maze Runner tests if a warning is emitted. I've raised [PLAT-12273](https://smartbear.atlassian.net/browse/PLAT-12273) to track that work

tomlongridge (approved):

PR #829, opened 2024-06-17
Only read Rack request body if it's rewindable

imjoehaines:
## Goal

Rack 3 made it possible for the request body to not be rewindable, meaning it can only be read once

We rely on reading the request body for two pieces of metadata, `request.params` (if it's a form request) and `request.body`. These won't be set if the body has already been read, which is quite likely

With `request.params` we will only read the query string if the request cannot be rewound

With `request.body` we will not read it at all if the request cannot be rewound

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Fixes
+
+* Only read Rack request body if it's rewindable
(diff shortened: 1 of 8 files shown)


tomlongridge (approved):
Should we update the docs to recommend rewindable (or state that non-rewindable will result in less data being captured)?

PR #835, opened 2025-06-03
Set default endpoints based on API key

twometresteve:
## Goal

Set default endpoints based on API key. If no endpoint is configured, payloads should be sent to *.insighthub.smartbear.com if the API key starts with `00000`.

## Design

Because the API key can be set in a block passed to `Bugsng.configure`, setting of the default endpoints needs to be moved later in the process.  

## Changeset

Endpoints are now set to `nil` in `Configuraation.initialize` and a new routine `set_default_endpoints` added to set them after any block has been called.

## Testing

Unit tests added and I've run some manual tests locally with an example app, verifying that the requested host is as expected for different API keys.

### DIFF
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -445,8 +445,8 @@ describe Bugsnag::Configuration do
           Bugsnag.configure do |config|
             config.api_key = '00000472bd130ac0ab0f52715bbdc600'
           end
-          expect(Bugsnag.configuration.endpoints.notify).to eq(Bugsnag::Configuration::HUB_NOTIFY)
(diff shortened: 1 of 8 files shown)


richardelms:
Looks ok to me 👍

spec/configuration_spec.rb
```diff
@@ -427,6 +431,23 @@ def output_lines
             '[Bugsnag] WARN: No valid API key has been set, notifications will not be sent'
           )
         end
+
+        it "uses the default endpoints for non-hub API keys", :no_configure => true do
+          Bugsnag.configure do |config|
+            config.api_key = '00002472bd130ac0ab0f52715bbdc600'
```
richardelms, line 437: clever

richardelms (approved):

PR #851, opened 2026-01-20
Guard against nil endpoints

agrzegorzewski:
## Goal

Under specific circumstances, the notifier might attempt delivery before the endpoint is configured. This PR adds logic to guard against this behavior.

## Changeset

- Added a check for the `url` being `nil`
- Added explicit handling of URI parse errors


twometresteve:
Please add an entry to the CHANGELOG, as a fix.

twometresteve (approved):

